PR #469, opened 2014-01-21
use dict headers for http request and response

ibizaman:
Following [this discussion](https://groups.google.com/forum/#!topic/elm-discuss/9AOFt2j72H8), I added headers handling on the Response side of the Http library.

Furthermore, I used a dictionary instead of list of tuples for the headers. I think this is more suited for what one would one to do with headers: search for the value associate with a header.

### DIFF
--- a/libraries/Http.elm
+++ b/libraries/Http.elm
@@ -16,32 +16,49 @@ you have very strict latency requirements.
 
 import open Signal
 import Native.Http
+import Dict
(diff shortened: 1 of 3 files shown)


evancz:
Alright, the new `Http` library in `evancz/elm-http` should handle this case!

Took a while to get the core API right, but I think we are there :)

PR #584, opened 2014-04-15
Add sendJsonp and map to Http module

osnr:
Open to suggestions. In particular, not sure about Http.map (what happened to #411?). Also not sure if we want to roll JSONP requests into Request, or have our own timeouts (JSONP can't return Failure right now, just an empty Success or stuck on Waiting). A bit of code duplication in Http.js too.

The approach I use for callback names and URL parameter is based on jQuery's.

Example of using JSONP for Wikipedia API lookups:
https://gist.github.com/osnr/10723233
(hosted at http://rsnous.com/demos/elm-jsonp/Jsonp.html)

### DIFF
--- a/libraries/Http.elm
+++ b/libraries/Http.elm
@@ -3,22 +3,33 @@ module Http where
 library if you have very strict latency requirements.
 
 # Sending Requests
-@docs send, sendGet
+@docs send, sendGet, sendJsonp
 
 # Creating Requests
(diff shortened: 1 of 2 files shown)


evancz:
The `evancz/elm-http` should handle this kind of thing now. Not sure if it makes sense to bring it in to that API though.

PR #591, opened 2014-04-21
Add find function to List module

rylev:
First of all, thanks to everyone for working on Elm! It's been a real pleasure working with it over the past few days. I'd love to contribute. To get my feet wet I decided to implement `List.find` which I needed several times on a pet project I'm working on. Let me know what you think.

### DIFF
--- a/libraries/List.elm
+++ b/libraries/List.elm
@@ -108,6 +108,15 @@ scanl1 = Native.List.scanl1
 filter : (a -> Bool) -> [a] -> [a]
 filter = Native.List.filter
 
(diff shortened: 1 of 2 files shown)


evancz:
To make sure we can handle these ideas in a scalable way we are doing [this](https://github.com/elm-lang/core/blob/master/CONTRIBUTING.md#adding-new-functions) now. Going with that advice, the `find` function appears [here](http://package.elm-lang.org/packages/circuithub/elm-list-extra/3.1.1/List-Extra#find) if you need it!

PR #601, opened 2014-04-29
Add slider to Graphics.Input

osnr:
Not entirely sure how Input stuff works, so please correct me if I got anything wrong (esp. in Native).

Here is a somewhat more complicated test case where you can edit the upper bound of the slider live:

```haskell
import Graphics.Input as Input
import Graphics.Input.Field as Field
import String

ip = Input.input 0

mc : Input.Input Field.Content
mc = Input.input Field.noContent

field = Field.field Field.defaultStyle mc.handle id "Max" <~ mc.signal

s m v = let ma = 
          case String.toFloat m.string of
                    Just f -> f
                    Nothing -> 5
        in Input.slider ip.handle id -5 ma 0.5 v

main = (\x y z -> x `above` y `above` z)
     <~ field ~ (lift2 s mc.signal ip.signal) ~ lift asText ip.signal
```

### DIFF
--- a/libraries/Graphics/Input.elm
+++ b/libraries/Graphics/Input.elm
@@ -23,7 +23,7 @@ examples in this library, so just read on to get a better idea of how it works!
 To learn about text fields, see the
 [`Graphics.Input.Field`](Graphics-Input-Field) library.
 
-@docs button, customButton, checkbox, dropDown
+@docs button, customButton, checkbox, dropDown, slider
 
 # Clicks and Hovers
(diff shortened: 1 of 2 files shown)


evancz:
I'd recommend using `elm-html` with `Html.toElement` for now. We are building up a list of potential revisions to core in https://github.com/elm-lang/core/issues/322, and `Graphics.*` may be getting split out so we can evolve that idea more independently.

PR #642, opened 2014-06-04
Add man pages

alexander-b:


evancz:
These seem to be out of date, and I'm not sure how to maintain these things as things progress. I think I'd need a maintainer who can keep this up to date (i.e. not me) if we were to put it in, so I don't think we can do it for now.

PR #644, opened 2014-06-07
Added support for links with the target attribute

DiegoNolan:
I added a small feature that allows a target attribute on link tags.

An example:

```haskell
main : Element                                                                    
main = flow down                                                                  
        [ linkWithTarget "http://google.com" "_blank" <| plainText "Link With Target"
        , link "http://google.com" <| plainText "Link without target"             
        ]  
```

The top link when clicked will open a new tab or window, depending on the browser to google, while the bottom one will change the current page to google.

### DIFF
--- a/libraries/Graphics/Element.elm
+++ b/libraries/Graphics/Element.elm
@@ -50,6 +50,7 @@ type Properties = {
   opacity : Float,
   color   : Maybe Color,
   href    : String,
+  target  : String,
   tag     : String,
   hover   : (),
   click   : ()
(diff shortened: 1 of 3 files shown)


evancz:
Yow, sorry for the wait!!!

This sort of thing lives on `elm-lang/core` these days. I don't think we want to be accepting API changes into there right this moment, and it looks like Graphics.* may move to its own world so it can be unlocked from the core libraries https://github.com/elm-lang/core/issues/324

PR #705, opened 2014-07-25
Add utcOffset

mgold:
[Mailing list thread](https://groups.google.com/forum/#!topic/elm-discuss/yXgezemiQtw). A relatively small but useful signal for localization, etc. Helps make apps more robust to timezones and DST while offloading that nightmare to JS.

### DIFF
--- a/libraries/Native/Signal/Time.js
+++ b/libraries/Native/Signal/Time.js
@@ -57,6 +57,7 @@ Elm.Native.Time.make = function(elm) {
       every : every,
       delay : NS.delay,
       timestamp : NS.timestamp,
+      utcOffset : Signal.constant(new Date().getTimezoneOffset() * -60000),
       since : F2(since),
       toDate : function(t) { return new window.Date(t); },
       read   : read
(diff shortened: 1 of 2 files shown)


evancz:
Adding to https://github.com/elm-lang/core/issues/322

bowsersenior:
Coming up on the 3rd anniversary of this issue (see https://github.com/elm-lang/elm-compiler/pull/705) and all I see is a shell game of links to other issues. With all due respect to the fantastic work on Elm, UTC functionality is a critical missing feature. Why not add a small, simple solution which can be enhanced to be simpler and more elegant in the future? Otherwise, the message you are sending is that critical production features like UTC support can be delayed for years pending a perfect solution being developed. In the field of time zones, the essential complexity of the feature makes it highly unlikely that a slick solution will ever materialize (see http://blog.mojotech.com/the-complexity-of-time-data-programming/).

Since native code cannot be included in 3rd-party elm modules, there isn't an easy way to get UTC support through an add-on module either. I've been impressed by the balance of practicality and elegance in Elm, but this is a textbook case of what's wrong with the functional language community, which tends to obsess over academic purity and neglect practical concerns.

mgold:
Indeed. This could be a whitelisted library if there is concern about expanding core.

PR #877, opened 2015-01-10
In port conversion, wrap record fields in strings to prevent bad obfuscation.

laszlopandy:
With this test file:
```elm
type alias Rec = { abc: Int, efg: Int }

port testOutPort : Rec
port testOutPort = { abc = 1, efg = 2 }   

port testInPort : Rec
```

It makes this diff in output:
```diff
--- test.js	2015-01-10 22:45:22.000000000 +0100
+++ test2.js	2015-01-10 22:59:20.000000000 +0100
@@ -425,16 +425,16 @@
    var testInPort = _P.portIn("testInPort",
    function (v) {
       return typeof v === "object" && "abc" in v && "efg" in v ? {_: {}
-                                                                 ,abc: typeof v.abc === "number" ? v.abc : _U.badPort("a number",
-                                                                 v.abc)
-                                                                 ,efg: typeof v.efg === "number" ? v.efg : _U.badPort("a number",
-                                                                 v.efg)} : _U.badPort("an object with fields \'abc\', \'efg\'",
+                                                                 ,abc: typeof v["abc"] === "number" ? v["abc"] : _U.badPort("a number",
+                                                                 v["abc"])
+                                                                 ,efg: typeof v["efg"] === "number" ? v["efg"] : _U.badPort("a number",
+                                                                 v["efg"])} : _U.badPort("an object with fields \'abc\', \'efg\'",
       v);
    });
    var testOutPort = _P.portOut("testOutPort",
    function (v) {
-      return {abc: v.abc
-             ,efg: v.efg};
+      return {"abc": v.abc
+             ,"efg": v.efg};
    },
    {_: {},abc: 1,efg: 2});
    var Rec = F2(function (a,b) {
```

### DIFF
--- a/src/Generate/JavaScript/Helpers.hs
+++ b/src/Generate/JavaScript/Helpers.hs
@@ -42,6 +42,11 @@ prop name =
     PropId () (var name)
 
 
+propBracketRef :: Expression () -> String -> Expression ()
+propBracketRef expr prop =
+    BracketRef () expr (StringLit () prop)
+
+
 string = StringLit ()
 
 
--- a/src/Generate/JavaScript/Ports.hs
+++ b/src/Generate/JavaScript/Ports.hs
@@ -127,7 +127,7 @@ inc tipe x =
           where
             object = ObjectLit () $ (prop "_", ObjectLit () []) : keys
             keys = map convert fields
-            convert (f,t) = (prop f, inc t (DotRef () x (var f)))
+            convert (f,t) = (prop f, inc t (propBracketRef x f))
 
 
 incomingTuple :: [CanonicalType] -> Expression () -> Expression ()
@@ -214,4 +214,4 @@ out tipe x =
           ObjectLit () keys
           where
             keys = map convert fields
-            convert (f,t) = (PropId () (var f), out t (DotRef () x (var f)))
+            convert (f,t) = (PropString () f, out t (DotRef () x (var f)))


dnalot:
My general thought is obfuscation isn't [what Closure is good for](https://developers.google.com/closure/compiler/docs/api-tutorial3). UglijyJS is right for that. It'd be nice to get some free static analysis though. I've considered tackling adding output annotations for that. Or helping someone else.

evancz:
This makes sense to me! Something is wrong that I can't merge automatically, but I'll hopefully get a chance to take a look manually. Joey's TCE stuff is touching this code too, so I need to run through it all soon.

@dnalot, point is that things would go badly wrong if you used Closure without this.

dnalot:
That doesn't sound right though. 1) Specifically in other PRs, you can export protected namespaces. I have contacts on their team that can help. 2) It's the wrong way to use Closure. Although it obfuscates with the best of them, it's a waste of a power tool. Uglify is even more configurable in that regard. A much greater win would be outputting JSDoc annotations for DCE. 3) I wouldn't think a potential optimization phase shouldn't shape your output. What happens when you don't Closure? It's kind of a heavy dependency anyway. Now you have a bunch of premature optimizations. If anything, add an phase to code generation. Iterate your IR (whatever it is in memory) one more time, add JSDoc and strings for prod builds.

dnalot:
Also do we have examples of failed Closure compiles? Are we passing the right flags?

dnalot:
I just confirmed. There are flags for renaming fields. Moreover we know all the API names and can export them as externs. Again Uglify is a better tool for this. But if you're willing to output JSDoc you can get respectable DCE until you implement it in house.

evancz:
Maybe still needed in new version, not sure.

PR #879, opened 2015-01-11
Expose modules even with obfuscation (ie. Elm['Main'] = {})

laszlopandy:
With this test file:
```elm
type alias Rec = { abc: Int, efg: Int }

port testOutPort : Rec
port testOutPort = { abc = 1, efg = 2 }   

port testInPort : Rec
```

This PR creates this diff:
```diff
--- test.js	2015-01-11 09:56:01.000000000 +0100
+++ test3.js	2015-01-11 10:29:20.000000000 +0100
@@ -1,5 +1,5 @@
var Elm = Elm || { Native: {} };
-Elm.Basics = Elm.Basics || {};
+Elm.Basics = Elm["Basics"] = Elm.Basics || {};
 Elm.Basics.make = function (_elm) {
    "use strict";
    _elm.Basics = _elm.Basics || {};
@@ -178,7 +178,7 @@
                         ,uncurry: uncurry};
    return _elm.Basics.values;
 };
-Elm.List = Elm.List || {};
+Elm.List = Elm["List"] = Elm.List || {};
 Elm.List.make = function (_elm) {
    "use strict";
    _elm.List = _elm.List || {};
@@ -410,7 +410,7 @@
                       ,sortWith: sortWith};
    return _elm.List.values;
 };
-Elm.Main = Elm.Main || {};
+Elm.Main = Elm["Main"] = Elm.Main || {};
 Elm.Main.make = function (_elm) {
    "use strict";
    _elm.Main = _elm.Main || {};
@@ -444,7 +444,7 @@
                       ,Rec: Rec};
    return _elm.Main.values;
 };
-Elm.Maybe = Elm.Maybe || {};
+Elm.Maybe = Elm["Maybe"] = Elm.Maybe || {};
 Elm.Maybe.make = function (_elm) {
    "use strict";
    _elm.Maybe = _elm.Maybe || {};
@@ -2305,7 +2305,7 @@
     };
 };
 
-Elm.Result = Elm.Result || {};
+Elm.Result = Elm["Result"] = Elm.Result || {};
 Elm.Result.make = function (_elm) {
    "use strict";
    _elm.Result = _elm.Result || {};
@@ -2550,7 +2550,7 @@
                         ,fromMaybe: fromMaybe};
    return _elm.Result.values;
 };
-Elm.Signal = Elm.Signal || {};
+Elm.Signal = Elm["Signal"] = Elm.Signal || {};
 Elm.Signal.make = function (_elm) {
    "use strict";
    _elm.Signal = _elm.Signal || {};

```

It is not necessary to expose & preserve the names of all the core modules, but there is no way to tell between core and user-defined ones.

### DIFF
--- a/src/Generate/JavaScript.hs
+++ b/src/Generate/JavaScript.hs
@@ -316,7 +316,7 @@ flattenLets defs lexpr@(A _ expr) =
 
 generate :: CanonicalModule -> String 
 generate modul =
-    show . prettyPrint $ setup "Elm" (names ++ ["make"]) ++
+    show . prettyPrint $ setup withExposedNamespaceInit "Elm" (names ++ ["make"]) ++
              [ assign ("Elm" : names ++ ["make"]) (function [localRuntime] programStmts) ]
   where
     names :: [String]
@@ -329,7 +329,7 @@ generate modul =
     programStmts =
         concat
         [ [ ExprStmt () (string "use strict") ]
-        , setup localRuntime (names ++ ["values"])
+        , setup withNamespaceInit localRuntime (names ++ ["values"])
         , [ IfSingleStmt () thisModule (ret thisModule) ]
         , [ VarDeclStmt () localVars ]
         , body
@@ -356,9 +356,17 @@ generate modul =
       where
         defs = mapM definition . fst . flattenLets [] $ Module.program (Module.body modul)
 
-    setup namespace path = map create paths
+    withNamespaceInit path = (InfixExpr () OpLOr (obj path) (ObjectLit () []))
+
+    withExposedNamespaceInit path =
+        AssignExpr () OpAssign exposedNs (withNamespaceInit path)
       where
-        create name = assign name (InfixExpr () OpLOr (obj name) (ObjectLit () []))
+          exposedNs = (LBracket () (bracketObj (init path)) (StringLit () (last path)))
+
+
+    setup makeRhs namespace path = map create paths
+      where
+        create name = assign name (makeRhs name)
         paths = drop 2 . init . List.inits $ namespace : path
 
     jsExports = assign (localRuntime : names ++ ["values"]) (ObjectLit () exs)
--- a/src/Generate/JavaScript/Helpers.hs
+++ b/src/Generate/JavaScript/Helpers.hs
@@ -51,6 +51,12 @@ obj vars =
       x:xs -> foldl (DotRef ()) (ref x) (map var xs)
       [] -> error "dotSep must be called on a non-empty list of variables"
 
+bracketObj :: [String] -> Expression ()
+bracketObj vars =
+    case vars of
+      x:xs -> foldl (BracketRef ()) (ref x) (map (StringLit ()) xs)
+      [] -> error "bracketObj must be called on a non-empty list of variables"
+
 
 -- Specific Utilities
 


cameron-martin:
Why `Elm.Main = Elm["Main"] = Elm.Main || {};`, rather than just `Elm["Main"] = Elm.Main || {};` or `Elm["Main"] = Elm["Main"] || {};`?

cameron-martin:
Wait, I think I'm missing the point of this PR altogether; Does closure compiler not treat `Elm["Main"]` the same as `Elm.Main`, even though they both do the same thing: access the `Main` property of the `Elm` object?

mgold:
I _think_ the point of the PR is that if you run the code through a JS minifier, the old object fields will be minified but the string literals will be untouched. This allows third-party code to still find the correct modules. Maybe.

cameron-martin:
Ah yeah, that makes sense.

dnalot:
@laszlopandy I think this might be a bad idea. Closure should have a setting to protect globals like `Elm`. I think we should fundamentally _not_ target Closure as a deploy target unless we intend to accept it as a dependency. That would drastically affect the shape of transpiled code including type annotations.

cameron-martin:
This PR actually was created with the aim to [bring closure compiler in as a dependency][1]. I'm not sure whether this is a good idea or not though.

[1]: https://groups.google.com/forum/#!topic/elm-dev/t_GfUjwI0bM

dnalot:
I think it's a _bad idea_<sup>:tm:</sup>. I foresee three issues:

0. One of my best friend's leads  development for a transpiler at Google. Although Closure is very effective at optimizing imperative code, in practice his output is shaped according to what Closure _can_ optimize rather than how he would necessarily structure it otherwise. Even then he has to walk to their desk and ask to fix stuff for weekly. We don't have that kind of access and the community is ripe for a project to replace it.  You don't want to have to rearchitect your compiler back-end when that happens. A post optimization phase wouldn't be a bad thing as long as it doesn't dictate unoptimized JavaScript especially if that JavaScript included ES6+ type annotations.
0. If you were to incorporate Closure, this is the wrong change. This should be accomplished with flags not in generated code. The biggest impact would be achieved using [JSDoc types annotations](https://developers.google.com/closure/compiler/docs/js-for-compiler).
0. Because Elm is relatively pure, there are an entire class of optimizations Closure simply cannot reason about. If you optimize output for Closure now, your input will change again when those optimizations are in place.

Have you already started a topic in [elm-dev](https://groups.google.com/forum/#!forum/elm-dev)?

dnalot:
A similar issue would be trying to optimize for V8. Thar be dragons :dragon:

cameron-martin:
I agree that structuring the output to please Closure Compiler would be the wrong approach.

The way I see it, whether JSDoc type annotations are a good fit here boils down to: Should generated code be documented? Instinctively I would say no, but I don't have anything else to add here.

Your third point suggests that it would be best to handle these kind of optimisations in-house. The three most easiest and most effective ways of minimising the size of generated code would probably be:

1. Whitespace elimination
2. Variable renaming
3. Dead code elimination

Out of interest, which project does your friend work on?

dnalot:
> structuring the output to please Closure Compiler would be the wrong approach.

:+1: 

> Should generated code be documented?

I actually think this would be a worthwhile post-optimization phase. Janis recently posted about [improving phases](https://groups.google.com/forum/?pli=1#!topic/elm-dev/3dmANqyxOlU) and this might be a good use case.

> 0. Whitespace elimination
> 0. Variable renaming
> 0. Dead code elimination

The former two could be offloaded to another dependency (esp. Uglify or Closure) unless the final transformation is optional and simple enough to implement unambiguously. The latter could be offloaded to Closure but I suspect we could do better with a more reason-able AST. I see the quickest path being annotated Closure short-term and in-house transformations long-term and perhaps Uglify.

> which project does your friend work on?

[GWT](https://twitter.com/gwtcreate/status/565812659044753408)

Would you mind starting a discussion in elm-dev and reference this discussion? I bet they'd have valuable input!

evancz:
@JoeyEremondi's work on dead code elimination will mean that we no longer use the same `Elm.Graphics.Element` structure in generated code. Instead I don't think we will have boundaries exactly like that and we will use canonical names, something like `elm-lang$core$2_0_1$Graphics$Element$show`. I suspect this will mean we can compress variables better.

About @laszlopandy's concern that the code does not need to be generated in one big chunk every time, I think this new generation strategy does not preclude that. We can still generate a single JS file for `Graphics.Element` but it'll just define things with canonical names and not filter anything out.

PR #892, opened 2015-02-10
Fix pretty printing of cons

JoeyEremondi:
Right now, cons expressions don't pretty print correctly due to incorrect parentheses. This is because the special check for :: is being done in Data expressions, when cons is being parsed as a BinOp.

This fix adds the check to the binOp case of the parentheses function.

### DIFF
--- a/src/AST/Expression/General.hs
+++ b/src/AST/Expression/General.hs
@@ -224,7 +224,7 @@ prettyParens (Annotation.A _ expr) =
   where
     needed =
       case expr of
-        Binop _ _ _ -> True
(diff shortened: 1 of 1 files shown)


evancz:
Can you show an example of when it messes up? How are you observing this?

Also, why the extra parens in the code fix?

JoeyEremondi:
Ah, another failed PR on my account.
I was giving an empty opTable in my example, which was the root of the problem.
I'll verify that it works with the proper associativity.

evancz:
So can this be closed?

PR #902, opened 2015-03-21
fix install script url

sonnym:
Nice and simple.  Not really much more to say!


evancz:
Cool, thank you!

sonnym:
De nada!  And, really, thank you for the wonderful language!

PR #907, opened 2015-04-03
Fix to reflect renamed type for JavaScript values through ports

ThomasWeiser:
Core module _Json.Encode_ is renamed to _JavaScript.Encode_

So the test for JavaScript values going through ports must be against type `JavaScript.Encode.Value`

Also renamed the test function itself to reflect its meaning more closely.

### DIFF
--- a/src/AST/Variable.hs
+++ b/src/AST/Variable.hs
@@ -61,9 +61,9 @@ is home name var =
     var == Canonical (Module home) name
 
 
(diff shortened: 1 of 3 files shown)


evancz:
Thanks for the PR! I actually came across this during a refactor I was doing yesterday, but hadn't gotten a chance to check it in yet. It should be fixed in master now!

PR #918, opened 2015-04-25
-Werror -> -Wall

maxsnew:
With more recent versions of transformers, ErrorT is deprecated so this
is causing build failures.

Wall will still print all of the same errors but won't stop installation
from working.

### DIFF
--- a/elm-compiler.cabal
+++ b/elm-compiler.cabal
@@ -33,7 +33,7 @@ source-repository head
 
 Library
     ghc-options:
-        -threaded -O2 -W -Werror
+        -threaded -O2 -W -Wall
 
     Hs-Source-Dirs:
         src


evancz:
I am in the process of converting to ExceptT (it is nice to get rid of that weird `noMsg` stuff) so that aspect of this will go away.

I like using -Werror for development and -Wall reports lots of additional errors that I do not care about, so these will not report the same errors or catch as many issues. Furthermore, you cannot upload to hackage with a -Werror flag, so no one going through cabal gets this. Finally, I think distributing binaries through npm will be the default pretty soon. Overall, I don't see the big win here.

evancz:
When you say "build failures" what do you mean? For testing, for people on certain platforms?

maxsnew:
Ah, of course -Wall is not the non-error version of -Werror at all, this pr
should just be removing -Werror.

I'm getting build errors on Travis because of this:
https://travis-ci.org/maxsnew/IO/builds/60021987




-Max Stewart New

evancz:
I think something like this happened. Thanks for the PR!

PR #921, opened 2015-04-30
Use ExceptT instead of ErrorT and update for GHC 7.10

nathan:
I'm submitting this as one PR so you don't have to deal with the merge conflicts, but the changes are available separately on the [nathan:mtl-except](https://github.com/elm-lang/elm-compiler/compare/nathan:mtl-except) and [nathan:ghc710](https://github.com/elm-lang/elm-compiler/compare/nathan:ghc710) branches.


cdepillabout:
@nathan, I think the imports from `Control.Applicative` need to be left in so the code can continue compiling on ghc-7.6 and ghc-7.8.  

Maybe you could use CPP macros like described [here](http://edsko.net/2014/09/13/haskell-cpp-macros/) for checking whether or not it's being compiled with ghc-7.10.

evancz:
Everything has been updated in the warnings branch, which will get merged into master soon. I need to fix a couple more things first, but it's on the way!

PR #924, opened 2015-05-04
Exhaustiveness checking

imeckler:
Gives warnings for inexhaustive and redundant pattern matches.


evancz:
Great, I'm excited for this! The style changes make a huge difference as a reader and reviewer, something about [processing fluency](http://en.wikipedia.org/wiki/Processing_fluency) that comes from having seen it this way I guess.

In any case, I think I'll see real stuff only when I start merging it in. There is this module called [`Type.ExtraChecks`](https://github.com/elm-lang/elm-compiler/blob/warnings/src/Type/ExtraChecks.hs) that does some checks that also need to be after type inference. I wonder if your PR and this should be merged into another validation phase. In any case, things are looking good now, and I'll do some more exploration as I try to merge it in!

PR #941, opened 2015-05-20
Update helper functions in Test.Compiler to fit new version of Compiler.compile.

manleyjster:
In particular, this PR accomodates the following changes to Compiler.compile:
* New parameter isRoot :: Bool (introduced in [fb1a6d26f0](https://github.com/elm-lang/elm-compiler/blob/fb1a6d26f09f3e43fc45df9df82ec548bbb276e7/src/Elm/Compiler.hs))
* Return value of type `([Warning], Either [Error] (PublicModule.Interface, String))`, rather than `Either String (PublicModule.Interface, String)`.

Note: This fixes `Test.Compiler`, but not `Test.Property`.  I got the tests to run by:

1. Removing `Test.Property` and `Test.Property.Arbitrary` from elm-compiler.cabal.
2. Removing `import Test.Property` and the reference to `propertyTests` in [`Test.hs`](https://github.com/manleyjster/elm-compiler/blob/fix-compiler-tests/tests/Test.hs#L13).

Running `cabal test` then results in all of the tests passing except for two involving ports:
* [`PortsWithTypeAliases.elm`](https://github.com/manleyjster/elm-compiler/blob/fix-compiler-tests/tests/test-files/good/PortsWithTypeAliases.elm)
* [`Ports.elm`](https://github.com/manleyjster/elm-compiler/blob/fix-compiler-tests/tests/test-files/good/Ports.elm)

Both of these tests fail with syntax errors - maybe the source files haven't been updated for version 0.15?

### DIFF
--- a/tests/Test/Compiler.hs
+++ b/tests/Test/Compiler.hs
@@ -3,6 +3,7 @@ module Test.Compiler (compilerTests) where
 
 import qualified Data.Map as Map
 import Data.Traversable (traverse)
+import Data.Bifunctor (first, second)
 
 import System.FilePath ((</>))
 import System.FilePath.Find (find, (==?), extension)
@@ -13,6 +14,8 @@ import Test.HUnit (Assertion, assertFailure, assertBool)
 
 import qualified Elm.Compiler as Compiler
 import qualified Elm.Compiler.Module as Module
+
+type ErrorMessage = String
 
 compilerTests :: Test
 compilerTests =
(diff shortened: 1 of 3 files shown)


tests/Test/Compiler.hs
```diff
@@ -53,21 +52,21 @@ testIf handleResult filePaths =
   where
     setupTest filePath =
       do  source <- readFile filePath
-          let result = Compiler.compile "elm-lang" "core" source Map.empty
+          let result = Compiler.compile "elm-lang" "core" True source Map.empty
```
manleyjster: Not sure if calling Compiler.compile with `True` for the `isRoot` parameter is correct here. Can you confirm/deny, @evancz?

tests/Test/Compiler.hs
```diff
@@ -53,21 +52,21 @@ testIf handleResult filePaths =
...
 
-isSuccess :: Either String a -> Assertion
-isSuccess result =
+isSuccess :: ([Compiler.Warning], Either [Compiler.Error] a) -> Assertion
+isSuccess (_, result) =
     case result of
-      Right _ -> assertBool "" True
-      Left msg -> assertFailure msg
+      Right _     -> assertBool "" True
+      Left errors -> assertFailure $ (unlines $ map (Compiler.errorToString "" "") errors)
```
manleyjster: Is there a better way to convert the errors to strings? I'm not sure what to pass to `Compiler.errorToString`.

tests/Test/Compiler.hs
evancz: If a file `isRoot` it is allowed to have ports. If it is a dependency, it cannot have ports.

tests/Test/Compiler.hs
evancz: The function ([here](https://github.com/elm-lang/elm-compiler/blob/master/src/Elm/Compiler.hs#L90)) takes in the file path as the first argument and the source code for the second argument. It will fail if you do not provide the source code.

It looks like both of these things are available, but this code needs to be slightly reworked for those values to flow to this point.

You can also use `concatMap` instead of `unlines . map` because it adds the appropriate amount of spaces already.

evancz:
Thanks for working on this, it looks quite close!

tests/Test/Compiler.hs
manleyjster: Fixed in 37e4dfe.

manleyjster:
Should be good to go.

@evancz - Can you take a quick look at [`PortsWithTypeAliases.elm`](https://github.com/manleyjster/elm-compiler/blob/fix-compiler-tests/tests/test-files/good/PortsWithTypeAliases.elm) and [`Ports.elm`](https://github.com/manleyjster/elm-compiler/blob/fix-compiler-tests/tests/test-files/good/Ports.elm)? They're failing with syntax errors. For someone who's familiar with the new ports syntax, I imagine that it should be a super quick fix. I took a look in the Changelog at the way that ports syntax changed in 0.15, but I couldn't see how to fix those test files.

evancz:
Great, thank you!

I made a couple style changes, got rid of the new dependency in favor of something less fancy, and fixed the two port examples. Thanks for taking the time to fix this! :D

manleyjster:
Perfect - sounds good! Thanks for the quick responses.

PR #945, opened 2015-05-28
Explain path to set ELM_HOME to after npm install

jvoigtlaender:
To improve user experience in a situation like this: https://github.com/kevva/elm-bin/issues/13.

### DIFF
--- a/src/Elm/Utils.hs
+++ b/src/Elm/Utils.hs
@@ -128,6 +128,9 @@ errorNotFound name =
     , "      C:/Program Files/Elm Platform/" ++ Version.version ++ "/share"
     , "      C:/Program Files (x86)/Elm Platform/" ++ Version.version ++ "/share"
     , ""
+    , "If you installed using npm, you have to set ELM_HOME to a certain directory"
+    , "of the form .../node_modules/elm/share"
+    , ""
     , "If it seems like a more complex issue, please report it here:"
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
Merge this before releasing 0.15.1? Could forestall some questions about "Why does the debug mode of elm-reactor not work after I installed Elm via npm?" in the future.

jvoigtlaender:
May I push this again? It's a pity that people who installed via `npm` don't get a helpful message in case they forgot to set the `ELM_HOME` variable (while people who used the Windows or Mac installers do get such a message in case that variable is not set for some reason).

mgold:
+1 for a better onboarding.

evancz:
This seems reasonable I guess, but the original message was added at a time when the Windows installer actually was somewhat unreliable about setting ELM_HOME. That is resolved now, so I suspect this is dead code for people taking that route now.

This is definitely not the correct "fix" and I would not say that this addresses the root "onboarding" issue at all. How come this isn't being done in the npm installer now? I started an issue for this https://github.com/kevva/elm-bin/issues/22 detailing the tricks I needed to do (that'd work on mac and linux at least). This is the root issue. An installer that does not set this is messed up, it's not a problem for the compiler. It's fine to help in the meantime, but this is not fixed at all.

Issue #948, opened 2015-05-31
Disallow case statements on non-exported types


mgold:
Minimal example, outputs `"red"` in 0.15 and master:

````elm
import Graphics.Element exposing (show)
import Color exposing (..)

main = show <| case purple of
  red -> "red"
  purple -> "purple"
  _ -> "some other color"
  ````

The compiler rejects the qualified names with a somewhat mysterious `expecting upper case variable`. However, the union type representation of colors is not exported, which I think is why this program (1) should be invalid (2) erroneously always takes the first option. This becomes more of a problem for larger opaque types, like `Dict`s and things with a Native implementation.

It's not a regression and it's fairly obscure, but hopefully it can get fixed at some point. Until then, multi-way if and equality checks work fine.

**Comment by jvoigtlaender** [comment]
Note that your `main` function is equivalent to:
```elm
main = show <| case purple of
  _ -> "red"
```
because your `red` is just a variable, not a constructor name.

Also, the compiler is perfectly right to reject a program like
```elm
main = show <| case purple of
  Color.red -> "red"
```
because a pattern can contain constructors and variables, but `Color.red` is neither a constructor nor a variable.

So, no compiler fix needed here. It's just a misunderstanding on your part, I guess.

jvoigtlaender:
One thing, though: 

The error message  `expecting upper case variable` for
```elm
main = show <| case ... of
  Color.red -> ...
```
is indeed a bit unfortunate. Maybe it should instead be `expecting upper case constructor name`?

jvoigtlaender:
But the way this error message is generated in the parser voids the possibility of mentioning `constructor` there, since it turns out the same part is responsible for parsing other upper case things (like module names). So I'd suggest to replace `upper case variable` by `upper case name` in that error message.

jvoigtlaender:
See [this PR](https://github.com/elm-lang/elm-compiler/pull/949).

mgold:
Ah. (Don't file issues late at night.) I was expecting to do something like this:

````elm
import Graphics.Element exposing (show)

type Foo = Bar Int
myFoo = Bar 5

main = show <| case myFoo of
  Bar 3 -> "nope"
  Bar 5 -> "hooray!"
  _ -> "something else"
````

[...]

PR #949, opened 2015-05-31
Improved parse error messages

jvoigtlaender:
Previously, e.g., on a parse error for a record label, the parser would complain it expected a `lower case variable`, which might confuse the programmer. (Why is a record label a "variable"? It should be a "name".) Similarly, in the parse error for `case ... of Module.lowercasename -> ...` the parser would complain it expected an `upper case variable`, which again is confusing (see https://github.com/elm-lang/elm-compiler/issues/948). Instead, it would make more sense to complain that an `upper case name` was expected (a constructor is not a "variable", but it is a "name").

### DIFF
--- a/src/Parse/Helpers.hs
+++ b/src/Parse/Helpers.hs
@@ -68,12 +68,12 @@ var =
 
 lowVar :: IParser String
 lowVar =
-  makeVar lower <?> "a lower case variable"
+  makeVar lower <?> "a lower case name"
 
 
 capVar :: IParser String
 capVar =
-  makeVar upper <?> "an upper case variable"
+  makeVar upper <?> "an upper case name"
 
 
 qualifiedVar :: IParser String
@@ -88,7 +88,7 @@ rLabel = lowVar
 
 innerVarChar :: IParser Char
 innerVarChar =
-  alphaNum <|> char '_' <|> char '\'' <?> "more letters in this variable"
+  alphaNum <|> char '_' <|> char '\'' <?> "more letters in this name"
 
 
 makeVar :: IParser Char -> IParser String
--- a/src/Parse/Helpers.hs
+++ b/src/Parse/Helpers.hs
@@ -63,7 +63,7 @@ iParseWithTable sourceName table aParser input =
 
 var :: IParser String
 var =
-  makeVar (letter <|> char '_') <?> "a variable"
+  makeVar (letter <|> char '_') <?> "a name"
 
 
 lowVar :: IParser String
--- a/src/Parse/Module.hs
+++ b/src/Parse/Module.hs
@@ -78,7 +78,7 @@ import' =
 
 listing :: IParser a -> IParser (Var.Listing a)
 listing item =
-  expecting "a listing of values to expose, like (..)" $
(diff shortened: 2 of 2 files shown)


evancz:
Can you share some cases where it actually *is* a variable? Is it an acceptable degradation of those messages to improve these cases? In practice, are the errors mentioned here more likely than ones that actually would be related to variables?

evancz:
Also, let's add some cases to https://github.com/evancz/error-message-catalog to cover these situations.

jvoigtlaender:
It's difficult for me to provide such cases, because I am not sure what other people consider to be "variables". For example, in this code:
```elm
f x = 10
```
are `f` and `x` "variables", or is only `x` a "variable"? I think I have seen both: people calling `f` a (top-level) variable as well, and people calling `f` a function and only `x` a variable.

jvoigtlaender:
Okay, here is a case in which I managed to trigger the "expected lower case variable" error:
```elm
port Z : Int
port z = 5
```
This currently gives the message `unexpected "Z"; expecting whitespace, a newline or lower case variable`. Is `z`/`Z` really a "variable" here? Wouldn't this message also read better as `unexpected "Z"; expecting whitespace, a newline or lower case name`?

mgold:
Yup, that's definitely not a variable. My first thought was "identifier" but "name" is certainly friendlier. So +1.

jvoigtlaender:
Here is another case:
```elm
import Graphics.Element exposing (0)
```
currently gives the error message `expecting whitespace, a newline or listing of values (x,y,z)`. Here I would say that "values" is really wrong. What we expect there are "names". Because we could also list types there, not just functions and constants. But we don't consider types to be values.

jvoigtlaender:
Sorry, that revert was accidental.

jvoigtlaender:
In
```elm
type u a = Z
```
the error message currently is: `expecting whitespace, a newline, "alias" or upper case variable`. Again, "name" would make more sense to me than "variable", since the `U` in
```elm
type U a = Z
```
is not a variable.

I realize that all this does little to answer your question Evan, how the quality of error messages would be with "name" if the thing referred to by that name is *really* a variable. The problem is, I have problems triggering such errors. Because it seems that in situations where such an error could occur, actual errors are preempted by other messages. For example, if I make a mistake like writing ``4 `Basics.Min` 6`` instead of ``4 `Basics.min` 6``, where the `expected lower case variable` error could occur (because of accidental `Min` instead of `min`), what instead happens is that the error is `expecting whitespace or end of input` (because the `Min` was interpreted as a potential module qualifier). Similar in other situations.

So most of the evidence I can provide are only cases where using "name" instead of "variable" or "value" is more appropriate. And I can only hypothesize that in cases where the thing referred to is indeed something one would take to be a "variable", at least it wouldn't be too harmful to say "name" in the error message.

jvoigtlaender:
I've added the cases from the above comments to https://github.com/evancz/error-message-catalog.

evancz:
Great, thank you for showing these examples, this seems a lot nicer! And it seems like it's strictly an improvement, great find!

PR #954, opened 2015-06-01
Fix a missing space

rtfeldman:
**Before:**

```
Type Shrink.Shrinker has too fewarguments.
```

**After:**

```
Type Shrink.Shrinker has too few arguments.
```


evancz:
Nice, thank you!

PR #955, opened 2015-06-01
Replace "want?!" with "want?"

rtfeldman:
**Before:** COMPILER IS ANGRY AT YOU :angry: 

```
Naming multiple values 'dropZone' in a single let-expression makes
things ambiguous. When you say 'dropZone' which one do you want?!

39|           dropZone = makeSection' True (" ", [])
              ^^^^^^^^
Find all the values named 'dropZone' in this let-expression and
do some renaming. Make sure the names are distinct!
```

**After:** Compiler is here to help :smile: 
```
Naming multiple values 'dropZone' in a single let-expression makes
things ambiguous. When you say 'dropZone' which one do you want?

39|           dropZone = makeSection' True (" ", [])
              ^^^^^^^^
Find all the values named 'dropZone' in this let-expression and
do some renaming. Make sure the names are distinct!
```

### DIFF
--- a/src/Reporting/Error/Syntax.hs
+++ b/src/Reporting/Error/Syntax.hs
@@ -79,7 +79,7 @@ toReport dealiaser err =
         Report.simple
           "DUPLICATE DEFINITION"
           ("Naming multiple top-level values '" ++ name ++ "' makes things\n"
-            ++ "ambiguous. When you say '" ++ name ++ "' which one do you want?!"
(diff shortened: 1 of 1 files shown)


mgold:
I'd suggest axing the exclamation at the very end while we're at it. "..and do some renaming to ensure that all names are distinct" or something

Fresheyeball:
:+1:  Elm has the most polite compiler ever.

PR #960, opened 2015-06-03
Use backticks instead of single quotes for reporting code terms

rtfeldman:
The motivation behind this PR was that I ran into this error:

```
Could not find variable 'makeSection''.

Close matches include:
    makeSection
```

At first I ran off to figure out why `makeSection` was not in scope, then eventually came back and realized it was actually complaining about a variable called `makeSection'`. I missed it because of the single quotes around the variable name.

This PR changes such usages to backticks instead, like so:

```
Could not find variable `makeSection'`.

Close matches include:
    makeSection
```

Arguments for this change:

* Single quotes are often used at the end of variable names, so the current collision will come up often.
* Backticks are not allowed in variable names, so there is no potential for such a collision there.
* Backticks are used in Markdown to denote code snippets, which is also what they do in these cases.

### DIFF
--- a/src/CommandLineRouter.hs
+++ b/src/CommandLineRouter.hs
@@ -26,10 +26,10 @@ attemptToRun command args =
       case found of
         Nothing ->
           hPutStrLn stderr $
-            "Could not find command '" ++ command ++ "'. Maybe there is a typo?\n\n\
+            "Could not find command `" ++ command ++ "`. Maybe there is a typo?\n\n\
             \Default commands include:\n\n"
             ++ availableCommands ++
-            "\nWhen you try to run the command '" ++ command ++ "' we actually search for an\n\
+            "\nWhen you try to run the command `" ++ command ++ "` we actually search for an\n\
             \  executable named elm-" ++ command ++ ". Are you able to run elm-" ++ command ++ "?\n\
             \  Is it on your PATH?"
 
--- a/src/Elm/Utils.hs
+++ b/src/Elm/Utils.hs
@@ -87,7 +87,7 @@ unwrappedRun command args =
 missingExe :: String -> CommandError
 missingExe command =
   MissingExe $
-    "Could not find command '" ++ command ++ "'. Do you have it installed?\n\
+    "Could not find command `" ++ command ++ "`. Do you have it installed?\n\
     \    Can it be run from anywhere? Is it on your PATH?"
 
 
--- a/src/Generate/Cases.hs
+++ b/src/Generate/Cases.hs
@@ -258,5 +258,5 @@ matchMix vs branches def =
 
 noMatch :: String -> a
 noMatch name =
-    error $ "unexpected pattern in '" ++ name ++
-            "' function. Report a compiler issue to <https://github.com/elm-lang/Elm>."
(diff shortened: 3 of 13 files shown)


evancz:
Makes sense. I think it's a bit uglier, but I buy the reasoning behind it.

laszlopandy:
What if we use real unicode quotes? They are more beautiful, and also easier to read.
`Could not find variable “makeSection'”.`

evancz:
Oooohh, I really like that idea! Can someone do a PR for that?

rtfeldman:
That was actually my first thought, but then I figured it might break in certain environments where people might want to run a compiler. Maybe that's not worth worrying about though? Happy to make another PR if we think that's a safe way to go, because I agree they'd look better.

evancz:
What environments are these?

Does using the typical `"` seem plausible?

rtfeldman:
I honestly didn't have a specific environment in mind (old Windows-based consoles, maybe?) but rather thought it prudent to err on the side of "this will definitely solve the problem and will definitely be supported everywhere."

I think the typical `"` is still troublesome with the original motivating use case, depending on the font. On GitHub it looks okay, but I've definitely seen monospace fonts where consecutive single quotes and double quotes are hard to distinguish.

```
Could not find variable "makeSection'".

Close matches include:
    makeSection
```

Compared to:

```
Could not find variable “makeSection'”.

Close matches include:
    makeSection
```

...or the backticks version:

```
Could not find variable `makeSection'`.

Close matches include:
    makeSection
```

Having said that, it just occurred to me that French quotation marks are both SUPER clear and actually happen to be in ASCII as well:

```
Could not find variable «makeSection'».

Close matches include:
    makeSection
```

Apanatshka:
French quotation is great, but that's in the *extended* ASCII table. It's
not standard across all machines. So that still doesn't work in some
non-Latin-1 environments...

rtfeldman:
Fair, but Unicode in general might actually be totally fine. I chose backticks to be conservative, but that may have been an irrelevant consideration.

I do like how clear the French quotation marks are, though. :smiley:

mgold:
Throwing in one more.

````
Could not find variable: makeSection'

Close matches include:
    makeSection
````

Hey, it's consistent.

**Comment by rtfeldman** [comment]
:open_mouth: 

Um, yeah, that's kind of a clear winner in my book.

I'd be happy to make another PR converting other such messages to use newlines to achieve a similar style.

**Comment by Fresheyeball** [comment]
This is a nit pick, but couldn't you have infixes with this issue? Infixes that use `:`...

```
Could not find variable: :+:

Close matches include:
    makeSection
```

dropping to a new line and indenting is the most clear so far, why not do it for both?

```
Could not find variable: 
    :+:

Close matches include:
    makeSection
```

rtfeldman:
Wow, I like that even better. Makes it easier to visually compare the
erroneous term with the suggested alternatives.

mgold:
Yeah, let's see it non-infix...
````
Could not find variable:
    makeSection'

Close matches include:
    makeSection
````

We have a winner.

rtfeldman:
@evancz if you're in agreement, I'll make a fresh PR.

PR #972, opened 2015-06-27
Add period to docs error

mgold:


evancz:
Thanks! When there's a typo like this, it's always easiest as a PR. You know exactly the context that it shows up in, so it's easy to find and do a PR!

mgold:
I mean, I sent [the link](https://github.com/elm-lang/elm-compiler/blob/a4e77b1b424db187db1496c3c290c1a76be0da7a/src/Reporting/Error/Docs.hs#L48), and it's not hard to search for string literals. But however you want to do it is fine.

PR #974, opened 2015-06-30
Change error message from 'a record' to 'this record'

rtfeldman:
Now that compiler error messages provide such clear context, it feels weird that the "field missing" error doesn't acknowledge that you already know which record is the broken one.

**Before:**

```
The 1st argument to function `isOver50` has an unexpected type.

9|          isOver50 hermann
                     ^^^^^^^
Looks like a record is missing the field `age`
```

**After:** ("a record" became "this record" in the last line)

```
The 1st argument to function `isOver50` has an unexpected type.

9|          isOver50 hermann
                     ^^^^^^^
Looks like this record is missing the field `age`
```

### DIFF
--- a/src/Type/Unify.hs
+++ b/src/Type/Unify.hs
@@ -364,9 +364,9 @@ fieldMismatchError missingFields =
     case Map.keys missingFields of
       [] -> ""
       [key] ->
-        "Looks like a record is missing the field `" ++ key ++ "`"
+        "Looks like this record is missing the field `" ++ key ++ "`"
       keys ->
-        "Looks like a record is missing fields "
+        "Looks like this record is missing fields "
         ++ List.intercalate ", " (init keys) ++ ", and " ++ last keys
 
 


evancz:
In this case, we know it's this specific record. The concern is that this error could potentially be triggered in other ways where the root cause is not *directly* the underlined thing. In that case, it seems like saying "it's this one" could be really really misleading. So maybe we can find some cases of that to confirm or deny this theory, and then weigh if it is worth it!

rtfeldman:
Gotcha. Make sense to me!

Apanatshka:
You're already saying it's *this* record by pointing to it. 
And at the same time by saying "Looks like" to make sure it doesn't look like a fact :)
I think this change doesn't retract from the "Looks like" which conveys the message that it may not be exactly where the compiler is pointing.

evancz:
Alright, without further investigation, I think it makes sense to close this. Let's just add examples like this to error-message-catalog so we build our our data set for now.

PR #977, opened 2015-07-02
Fix #962; Parsing record patterns in cases

### LINKED ISSUE #962: Parsing issue with record destructuring in case statement
As discussed on the mailing list: https://groups.google.com/d/topic/elm-discuss/w9fjpRwPZH4/discussion

Minimal example:

```
foo = case {x=1} of
  {x} -> x == 1
```

Expected behaviour: accepted as correct program. 
Actual behaviour: parse error on line of the record pattern `{x}`.

Tzbob:
This fixes the parsing issue in #962 by looking ahead when checking for one-liner syntax in cases, e.g.:

```case record of { {x} -> x; {y} -> y }```

It does not address issues mentioned with this kind of case expressions (warnings to switch  to let for example).

### CODE REVIEW

**Comment by evancz** [comment]
Having a `try` like this means you can get really bad backtracking that hides syntax errors. I think the risk of bad backtracking outweighs the benefit of closing that issue (especially since the example program you wrote will always take the first branch)

Perhaps the right move is to remove the `{}` syntax for `case` expressions and then it won't be ambiguous anymore.

**Comment by Tzbob** [comment]
Agreed, I'm not a fan of the { ; } syntax anyway. I'll modify the branch. 

Feel free to leave this pull request open if more changes to the { ; } syntax should be made.

**Comment by mgold** [comment]
So if I understand how this (including the parent feature) would work, it's
```elm
magnitude : {a | x : Float, y: Float} -> Float
magnitude record =
    case record of
        {x,y,z} -> sqrt <| x*x + y*y* + z*z
        {x,y} -> sqrt <| x*x + y*y
````
Which allows you to use the same function on different dimensions of vector. That's pretty neat, actually.

**Comment by Apanatshka** [comment]
@mgold this PR and the linked issue are about parsing. What you're talking about would be a type system feature, which would probably look slightly different from what you envision there. If you read the [paper about extensible records](http://research.microsoft.com/pubs/65409/scopedlabels.pdf) on which Elm's system is based, you may have noticed an extensible variants (union types) system in section 5. That would be a good option for modeling the features you're looking for.

**Comment by mgold** [comment]
@Apanatshka What I'm trying to understand is how a statement like `case record of { {x} -> x; {y} -> y }` could be valid if we know the exact field in the record. To me, that code would mean "if `record` has an `x` field, return it, or else if it has a `y` field, return it, otherwise fail at runtime with an incomplete pattern match, and also `record.x` and `record.y` must have the same type". That doesn't sound very useful to me, and doesn't sound possible without the type system feature from my previous post. So I guess I'm still confused as to what parsing issue is being solved.

**Comment by evancz** [comment]
The code snippet is kind of goofy. For `case record of { {x} -> x; {y} -> y }` to be well-typed, `record` must have an `x` and a `y` field. Furthermore, those fields must have the same type. So the first branch will always be taken in a well-typed program.

The example you gave is wrong in that the type would need to specify the `z` field and then the first branch would always be taken.

Overall, I don't see this code ever being written in "good code" but I think it's still a bug that you cannot put a record match in this position!

**Comment by mgold** [comment]
I'm confused. If we haven't been able to come up with a good example, and we don't think that that it makes for "good code", then why should we include it? From what I understand, it would let you write

```elm
case record of
    {x, y} -> f x y
```
instead of
```elm
let {x, y} = record
in f x y
```

This seems like an abuse of the `case` construct, in that any subsequent cases are always ignored (right?). This seems like a "gottcha", and we've already seen "the first case is always evaluated" can be [confusing to newcomers](https://groups.google.com/d/msg/elm-discuss/BCOwYRTm7uk/sfR-t5-zXsMJ).

Tzbob:
I agree with Evan, it's a bug that you cannot write this code. 

Regarding the 'gottcha' part -- I think it would be a lot more confusing if you cannot write all legitimate patterns in a case clause. It might not be good code but it's correct code like writing ```if true```, ```1 + 0``` etc.

I'd say remove the syntax error (like in this pull request) and add this to a table of code patterns that can/should trigger warnings.

PR #997, opened 2015-07-19
Use SVG for Travis CI badge

MaxGabriel:
If you look at the [rich diff](https://github.com/elm-lang/elm-compiler/pull/997/files?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8) it looks much nicer, especially on retina displays.


evancz:
Nice, did not know that was an option!

PR #1016, opened 2015-08-17
Relax bounds on blaze-... libraries

jvoigtlaender:
Are there specific reasons that `blaze-html` was constrained to `< 0.8` and `blaze-markup` to `< 0.7`?

These are the only constraints that currently keep this repository from building with the set of constraints given by "Long Term Support Haskell 3.1": https://www.stackage.org/lts-3.1/cabal.config


evancz:
We wanted to give upper and lower bounds because new versions kept breaking everything. If things build with the new stuff it's fine. Point was to make sure we test things before some maintainer can release a breaking `0.0.0.x` change as some folks do.

Issue #826, opened 2014-11-11
Module name collision in generated code


evancz:
As it is, each module is attached to Elm according to its name, so `Search.View` becomes `Elm.Search.View`.

This becomes problematic when there are many packages involved that may have modules with the same names. For example, if there's a private `Utils` module in `evancz/focus` and in `evancz/automaton`, the module that wins at runtime depends on what order things were compiled.

The solution is to use fully qualified names in generated code, so `Utils` would be `Elm.evancz.focus.Utils` or something like that.

joaalto:
I just ran into a module name collision. I'm using [rluiten/elm-date-extra](http://package.elm-lang.org/packages/rluiten/elm-date-extra/3.0.0) in an hour reporting app I'm writing. Then I needed some UI components and decided to bring in [Elm-UI](https://github.com/gdotdesign/elm-ui). Unfortunately this doesn't work, as Elm-UI depends on [mgold/elm-date-format](http://package.elm-lang.org/packages/mgold/elm-date-format/1.1.3/), and both `elm-date-format` and `elm-date-extra` have a module called `Date.Format`:

```
I found multiple modules named 'Date.Format'.

Module 'Ui.DatePicker' is trying to import it.

Modules with that name were found in the following locations:

    package mgold/elm-date-format
    package rluiten/elm-date-extra
```

evancz:
Whoo, resolved in 0.17!

szagi3891:
how can use it?

evancz:
To clarify, the initial issue was that if two packages have *private* modules with the same name, it used to cause problems. So if package `A` had a `Utils` module that **was not** exposed, and if package `B` had a `Utils` module that **was not** exposed, it would mess up.

Then someone posted about a separate issue, thinking this was tracking what he was worried about. That happens, and that is why I recommend always opening a *new* issue rather than piling on things that you think are your problem.

PR #1017, opened 2015-08-18
Add the source file and line number to docs entries

ajhager:
This PR adds an entry in the docs for each item with the source file and line number. The need for this is outlined in elm-lang/elm-make#29 and  elm-lang/package.elm-lang.org#35, but in summary it allows tools that consume the docs json file to link to the exact source file and line number where something is defined. This could be "go to definition" in editors or direct links in the package docs.

This PR also relies on a small PR to elm-lang/elm-make#45. Both of these PRs will fail on CI of course, because they depend on each other, so it will probably need to be tested manually.

### DIFF
--- a/src/Elm/Compiler.hs
+++ b/src/Elm/Compiler.hs
@@ -66,10 +66,11 @@ parseDependencies sourceCode =
 compile
     :: Context
     -> String
+    -> String
     -> Map.Map PublicModule.Name PublicModule.Interface
     -> (Dealiaser, [Warning], Either [Error] Result)
 
-compile context source interfaces =
+compile context path source interfaces =
   let
     (Context user packageName isRoot isExposed) =
       context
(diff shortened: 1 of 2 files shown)


evancz:
My feeling is that it makes sense to pass this information in with the `Context` value. We will be wanting the full package name to disambiguate modules as described in #826, and it feels like this fits nicely in that chunk of information.

The net goal here is to make "jump to source" work in the package catalog and in editors, correct? I guess this seems like a logical place to put this. I am assuming it'd come along with features like "tell me the type of this imported thing!" or "show me the docs for this imported thing!" so you'd need to be running the doc generator anyway. Is that fair to say?

ajhager:
I think putting into the context makes sense for sure; Good point. I will update my my PR. 

And yeah, the goal is to use this for "jump to source". And since we get all of the other information needed about a token from the docs files already, this tiny extra info enables opening the file under elm-stuff/packages or linking to the file on github.

evancz:
Cool, sounds good to me.

I just merged a bunch of stuff into master, so things are messed up for now. Let's do this as a separate PR because the stuff here probably got trampled!

PR #1023, opened 2015-08-24
Add benchmarks to elm-compiler

JoeyEremondi:
Add a benchmark suite to elm-compiler.

To test a particular elm version, just do
`elm make CompilerBenchmarks.elm --output=Benchmark.html` with whatever version of Elm you are testing, then load the resulting Html page in your browser.


Fresheyeball:
Wow! This is amazing!

evancz:
Awesome, very excited to have this, thank you! :D

PR #1026, opened 2015-08-25
Top level qualify

JoeyEremondi:
Introduces a new variant of `AST.Variable.Canonical`, namely `TopLevel`, which adds a module name to top-level local definitions.

This paves the road for lifting all outputted declarations to the top level.

For testing, I've ensured that this successfully builds several packages, including all of core. Additionally, JS output is unchanged for the entire Benchmark suite.

### DIFF
--- a/bench/CompilerBenchmarks.elm
+++ b/bench/CompilerBenchmarks.elm
@@ -1,0 +1,165 @@
+import Graphics.Element exposing (show)
+import Signal
+import Benchmark
+import Task exposing (Task, andThen)
+import Text
+import Array
+import Random
+import Dict
+import Time
+
+import LargeDictionary
+import ArrayBench
+import ListBench
+
+import RecordBench
+import SmallRecordBench
+import TupleBench
+import SmallTupleBench
+import CtorBench
+import SmallCtorBench
+import PatternMatchBench
+import IfBench
+import DictBench
+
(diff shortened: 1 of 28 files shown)


evancz:
This is looking good! Surprisingly minor changes. I'm gonna try to merge it in soon.

evancz:
Okay, I realized what was weird with the `[v1, v2]` part, and it just was something that surprised me in the existing code.

I had imagined that when you defined a variable in some pattern, it would take over everything. The imported things or built in things, none of that would be accessible in the environment anymore. Once it has been shadowed, it is not available anymore.

I guess I implemented it in a freaky way, and it happens to also work. I don't know if I knew I was doing this freaky thing when I did it.

In any case, that is why I was surprised by the `[v1,v2]` part. My mental model did not match the actual code.

That said, I think there is still a bug there. If a `Local` and a `TopLevel` are available, the `Local` should always win. I kinda want to just fix the underlying weirdness, so let me look into this a bit more for now.

JoeyEremondi:
>
> If a Local and a TopLevel are available, the Local should always win.


That's the whole point of the [v1, v2]. It checks if both a local and a
top-level are in the environment, and prefers the local one.

evancz:
Ah, I read read the body wrong, gotcha!

PR #1028, opened 2015-08-26
Fixes cabal file to work with stack

ianbollinger:
`stack` could not resolve the needed build artifacts without adding `elm-compiler` as a dependency to the corresponding executable. Additionally, this bumps the upper bounds on `blaze-html` and `blaze-markup` and adds `Paths_elm_compiler` to `other-modules` in order to silence a warning from `stack`.

### DIFF
--- a/elm-compiler.cabal
+++ b/elm-compiler.cabal
@@ -160,11 +160,13 @@ Executable elm
         CommandLineRouter.hs
 
     other-modules:
-        Elm.Compiler.Version
+        Elm.Compiler.Version,
+        Paths_elm_compiler
 
     Build-depends:
         base >=4.2 && <5,
         directory >= 1.0 && < 2.0,
+        elm-compiler,
         process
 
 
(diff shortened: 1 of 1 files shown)


elm-compiler.cabal
```diff
@@ -160,11 +160,13 @@ Executable elm
...
 
     other-modules:
-        Elm.Compiler.Version
+        Elm.Compiler.Version,
+        Paths_elm_compiler
 
     Build-depends:
         base >=4.2 && <5,
         directory >= 1.0 && < 2.0,
+        elm-compiler,
```
evancz: Why do we need to add this? The code does not depend on elm-compiler. It uses Elm.Compiler.Version [here](https://github.com/elm-lang/elm-compiler/blob/master/src/CommandLineRouter.hs#L9), so I get the transitive other-modules addition. It is not clear why that requires the elm-compiler dependency though, or what that would even mean.

elm-compiler.cabal
```diff
@@ -197,8 +199,8 @@ Test-Suite compiler-tests
         ansi-terminal >= 0.6.2.1 && < 0.7,
         base >=4.2 && <5,
         binary >= 0.7.0.0 && < 0.8,
-        blaze-html >= 0.5 && < 0.8,
-        blaze-markup >= 0.5.1 && < 0.7,
+        blaze-html >= 0.5 && < 0.9,
+        blaze-markup >= 0.5.1 && < 0.8,
```
evancz: Is this just lining things up with other things in this file? If so, can you link to the line numbers?

elm-compiler.cabal
ianbollinger: For some reason stack seems to fail unless an executable requires the corresponding library in every case. It appears to be more of a bug than a design decision, though it could be both.

evancz: Do you mind opening an issue with them to ask about it? If it is a bug, I'd be happy to remove that line!

PR #1032, opened 2015-08-27
Common package module (take 2)

JoeyEremondi:
SEE ALSO: related PRs for [elm-make](https://github.com/elm-lang/elm-make/pull/52) and [elm-package](https://github.com/elm-lang/elm-package/pull/127)

Refactors so that the `Name` type is in elm-compiler, so that we can pass Package names around the compiler process.

We create a new module, `Elm.Package`, containing the `Name` and `Version` types for packages.

`toString :: Version -> String` was renamed `verstionToString`, to avoid conflicts with `toString :: Name -> String`. Similarly with `fromString`.

We've moved the `elm :: Version` from Elm.Package.Version to Elm.Compiler, and renamed it to `version :: Version`.

### DIFF
--- a/elm-compiler.cabal
+++ b/elm-compiler.cabal
@@ -41,7 +41,9 @@ Library
         Elm.Compiler,
         Elm.Compiler.Module,
         Elm.Compiler.Type,
+        Elm.Compiler.Version,
         Elm.Docs,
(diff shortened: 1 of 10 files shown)


evancz:
This looks awesome!

PR #1033, opened 2015-08-28
Avoid conflicts when modules have the same name

JoeyEremondi:
SEE ALSO: [PR in elm-make](https://github.com/elm-lang/elm-make/pull/53).

This makes the changes necessary so that, when two packages have hidden modules with the same name, problems don't arise.

The ModuleNames we receive from Elm-make are now qualified by package name, and version. We also receive a dictionary mapping module names to package names for each imported package, so we know which interface files to give `Canonicalize.module'`. This dictionary has been added to the `Elm.Compiler.Context` type.

In Elm.Compiler.Module, we've added the `CanonicalName` type, which qualifies module names by package name and version. A similar type is in AST.Module, which uses AST.Module.Name instead of Elm.Module.Name.

The `Package` tuple synonym has been added to Elm.Package, which represents complete package information (name and version).

### DIFF
--- a/src/AST/Module.hs
+++ b/src/AST/Module.hs
@@ -1,11 +1,12 @@
 module AST.Module
-    ( Interfaces
+    ( Interfaces, CanonicalInterfaces
     , Types, Aliases, ADTs
     , AdtInfo, CanonicalAdt
     , SourceModule, ValidModule, CanonicalModule, Optimized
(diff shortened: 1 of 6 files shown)


JoeyEremondi:
Need to fix elm-compiler. Sorry for the noise.

PR #1038, opened 2015-08-29
Add Facts to Canonical.Def

JoeyEremondi:
Adds annotations containing the referenced top-levels and imported variables for each Definition in a module.


evancz:
Okay, should be all set for the `Sort` and DCE changes now!

I'm gonna leave `Generate.JavaScript.generate` undefined for now. I want to do a bit of reshuffling tonight though, so be sure to get a fresh version next time you do stuff!

PR #1048, opened 2015-09-06
Test fix empty expected files

laszlopandy:
Right now Generate/JavaScript.hs has a lot of errors. So if you run:
`ELM_WRITE_NEW_EXPECTED=1 cabal test`
It will write empty files for all the examples with errors.

But there are several other test cases which do compile and do need to be updated, and you don't want to have to pick through which ones to commit.
@evancz


evancz:
I don't fully understand, but this seems like an okay thing to add.

PR #1053, opened 2015-09-13
Migrate to Travis-CI container-based infrastructure

avh4:
http://docs.travis-ci.com/user/migrating-from-legacy/

This updates the Travis-CI script to use the new container-based infrastructure, which supposedly will lead to faster builds and more resources being allocated to the builds.  Sadly, the travis.yml is a bit uglier and has some duplication to support this.


evancz:
This looks good to me. Since GHC before 7.10 is going to stop working on the next version of OSX, I made it so 7.10 was the only version we tried to be compatible with. If this PR goes that route, it should be all set!

evancz:
As an aside, does this make things faster? It is able to grab this stuff off the shelf or something?

avh4:
Updated to only build with ghc 7.10.  I did a merge instead of a rebase so that the example of doing a build matrix will sill be in the git history.

I didn't notice a much of a difference in the build times with this, but supposedly the biggest speed increase on the new infrastructure is that builds will sit in the queue for less time when travis is under heavy load.  Also, I expect travis will start making incentives to phase out the legacy infrastructure at some point.

Another touted advantage is that you can cache dependencies (the cabal dependencies in our case), but I haven't taken a look at whether that's possible for us to use.

evancz:
Awesome, thanks for figuring this out!

PR #1057, opened 2015-09-15
Enable caching of ~/.cabal on Travis-CI

avh4:
Enable caching of ~/.cabal on Travis-CI

Depends on #1053

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 sudo: false
 
+cache:
+  directories:
+  - $HOME/.cabal
+
 addons:
   apt:
     sources:


avh4:
(It is not to me yet whether I am caching the correct directory, and if so whether caching it will actually speed anything up for the build...   Don't merge this until we've verified all that.)

laszlopandy:
I'm not sure if that will cache the modules from elm-compiler too. That
would be a bad thing. We want a clean build each time.

avh4:
This apparently didn't work.  On a subsequent build, Travis reported that it found the cache, but the build look just as long:  either ~/.cabal is not the correct folder to cache, or caching that folder has no effect on `cabal install --only-dependencies --enable-tests`

evancz:
Okay, it sounds like we don't want to do this now, but it'd be excellent if we figure it out someday :)

PR #1064, opened 2015-09-20
Add MultiParamTypeClasses pragma where appropriate

avh4:
This allows the elm-compiler source code to be properly parsed by haskell-src-exts.  (This allows, for instance, elm-compiler to be processed by [SourceGraph](https://hackage.haskell.org/package/SourceGraph))

For background: GHC has a bug that allows successful compilation without the MultiParamTypeClasses pragma, but such files are "not valid standard haskell".  See https://github.com/haskell-suite/haskell-src-exts/issues/29


evancz:
So this is needed anytime I use `FlexibleContexts`?

Issue #1072, opened 2015-09-30
Unit types not printed properly for documentation


jvoigtlaender:
A symptom of this is that http://package.elm-lang.org/packages/evancz/elm-effects/1.0.0/Effects#toTask shows the following syntactically incorrect type signature:
```elm
toTask : Address a -> Effects a -> Task Never )
```
An intermediate cause is that already the `.json` generated by the compiler for documentation purposes is wrong: http://package.elm-lang.org/packages/evancz/elm-effects/1.0.0/docs/Effects.json contains that wrong signature as well.

Since most of the packages on http://package.elm-lang.org/ will have to be updated to new versions with Elm 0.16, and since this means that a lot of documentation will be newly generated (and will then stick around for a while), it would be good to fix this error before the 0.16 release.

I already did some analysis of where the error could be, [over here](https://github.com/elm-lang/package.elm-lang.org/issues/66#issuecomment-130408404), but it has not been taken further yet.

mgold:
Wasn't I part of a fix for this once? Anyway, it's not for all types, see [Mouse.clicks](http://package.elm-lang.org/packages/elm-lang/core/2.1.0/Mouse#clicks).

jvoigtlaender:
Wrong. Read [over there](https://github.com/elm-lang/package.elm-lang.org/issues/66#issuecomment-130411473) and you will see why the fact that it looks okay in current documentation of `Mouse.clicks` is irrelevant. (Reason: The `.json` for that was generated long time ago, with `"generated-with-elm-version": "old"`. But what is problematic is that stuff generated with `"generated-with-elm-version":"0.15.1"` is incorrect.)

jvoigtlaender:
And an update: The problem does *not* seem to come from any of the functions I linked to [here](https://github.com/elm-lang/package.elm-lang.org/issues/66#issuecomment-130408404). The `prettyJson` from https://github.com/elm-lang/elm-compiler/blob/0b499e6befeba12ef1543c2e79f40299bda813a6/src/Elm/Docs.hs#L85 does return `"()"` for input `Type "_Tuple0"`.

mgold:
Is this the only file affected? How hard would it be to fix it manually (since it's "live in production")? Obviously we'd need to track down the root cause as well.

jvoigtlaender:
Obviously it's not the only file affected, given that I have already provided two examples (here and in the other comments I asked you to read for context):
[...]

PR #1075, opened 2015-10-02
Fix pretty-printing bug concerning unit type

jvoigtlaender:
Should fix https://github.com/elm-lang/elm-compiler/issues/1072.


evancz:
Ahh, very sneaky! Great find!

PR #1091, opened 2015-10-09
Inline to fix shadowing

### LINKED ISSUE #1086: "Cannot read property 'ctor' of undefined" on 0.16-alpha
The following code produces a `Cannot read property 'ctor' of undefined`

```elm
import Graphics.Element exposing (show) 

type AType = First     (() -> Bool)
               | Second    (() -> Bool) String String

getSecond : a -> a -> AType
getSecond a b = Second (\_ -> a == b) (toString a) (toString b)

toName : AType -> String
toName a =
    let 
        name = case a of
                 First _ -> "True"
                 Second _ a b    -> a ++ " == " ++ b
    in 
        name 

main = show <| toName <| getSecond  1 1
```

If I get rid of the let, the error disappears:
```elm
import Graphics.Element exposing (show) 

type AType = First     (() -> Bool)
               | Second    (() -> Bool) String String

getSecond : a -> a -> AType
getSecond a b = Second (\_ -> a == b) (toString a) (toString b)

toName : AType -> String
toName a =
    case a of
        First _ -> "True"
        Second _ a b    -> a ++ " == " ++ b

main = show <| toName <| getSecond  1 1
```

Adding a quote to variable `a` as in `a'` also gets rid of the error 
```elm
import Graphics.Element exposing (show) 

type AType = First     (() -> Bool)
               | Second    (() -> Bool) String String

getSecond : a -> a -> AType
getSecond a b = Second (\_ -> a == b) (toString a) (toString b)

toName : AType -> String
toName a =
    let 
        name = case a of
                 First _ -> "True"
                 Second _ a' b    -> a' ++ " == " ++ b
    in 
        name 

main = show <| toName <| getSecond  1 1
```

If only one case remains, the error becomes `Cannot read property '_1' of undefined`
```elm
import Graphics.Element exposing (show) 

type AType = Second    (() -> Bool) String String

getSecond : a -> a -> AType
getSecond a b = Second (\_ -> a == b) (toString a) (toString b)

toName : AType -> String
toName a =
    let 
        name = case a of
                 Second _ a b    -> a ++ " == " ++ b
    in 
        name 

main = show <| toName <| getSecond  1 1
```

### LINKED ISSUE #1090: Shadowing variable in case branch may lead to broken javascript code
When you write something like [the following](http://share-elm.com/sprout/5616dc31e4b0966c193dd419)
```elm
foo : Maybe number -> number
foo variableInCase =
  case variableInCase of
    Just n ->
      let
        variableInCase = Just "anything you want"  -- shadow variableInCase used in case
      in
        n  -- which will be used here in the javascript code to extract the field (variableInCase._0)

    _ ->
      0
```
elm will compile the ```n``` in the let body to ```variableInCase._0```, but which has been shadowed and the result is now the string ```"anything you want"```.

Edit: on elm 0.15

evancz:
Fix #1086 and fix #1090

### DIFF
--- a/elm-compiler.cabal
+++ b/elm-compiler.cabal
@@ -88,6 +88,7 @@ Library
         Nitpick.TopLevelTypes,
         Optimize,
         Optimize.Environment,
+        Optimize.Inline,
         Optimize.Patterns,
         Optimize.Patterns.DecisionTree,
         Parse.Binop,
--- a/src/AST/Expression/Optimized.hs
+++ b/src/AST/Expression/Optimized.hs
@@ -1,7 +1,8 @@
 {-# OPTIONS_GHC -Wall #-}
 module AST.Expression.Optimized
     ( Def(..), Facts(..), dummyFacts
-    , Expr(..), Decision(..)
+    , Expr(..)
+    , Decider(..), Choice(..)
     ) where
 
 import qualified AST.Expression.General as General
@@ -19,14 +20,12 @@ import qualified Reporting.Region as R
 data Def
     = Def Facts String Expr
     | TailDef Facts String [String] Expr
-    deriving (Eq)
 
 
 data Facts = Facts
     { home :: Maybe ModuleName.Canonical
     , dependencies :: [Var.TopLevel]
     }
-    deriving (Eq)
 
 
 dummyFacts :: Facts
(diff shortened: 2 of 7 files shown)


laszlopandy:
Are there any example Elm files in the good/ directory which test the new expected output? If not one should be added to avoid regressions.

evancz:
Do you mean tests that we run to see if they have any issues? Or do you mean tests where we look at the generated JS to make sure it is what we'd expect?

laszlopandy:
The generated JS tests, I mean.

PR #1098, opened 2015-10-10
Check in old benchmark files so we can race 0.15 and 0.16

evancz:


mgold:
Interested to see the results :checkered_flag:

evancz:
[They are very good](https://groups.google.com/forum/#!topic/elm-dev/m5WgdwSms5I) in Chrome and Opera. It seems to be less amazing in other browsers.

PR #1101, opened 2015-10-11
Whitespace fixes

evancz:


evancz:
@avh4, this stuff might be interesting for you. I think it'll be a bit faster this way at the very least.

avh4:
Ah, good!  It looks simpler.

PR #1118, opened 2015-10-23
Add Debug to default imports, fix #1107 and get Debug.crash working again

### LINKED ISSUE #1107: Generate good code for Debug.crash
Right now it is broken. It should be not broken.

evancz:


mgold:
May I ask the rationale for this? The default imports have been kept small, and in theory you shouldn't use the things in Debug.

jvoigtlaender:
@mgold, see the thread on elm-dev (first search result for "Debug").

mgold:
Okay thanks!

PR #1129, opened 2015-10-27
Tell users which repo they should report an error to

jvoigtlaender:
We see reporting issues to the "wrong" repos again and again. So better be explicit.


mgold:
:+1:

evancz:
Thanks! Had to update it for master, so I added a bit of extra info as well at https://github.com/elm-lang/elm-compiler/pull/1133

PR #1131, opened 2015-10-28
Add hints for incomplete pattern matches

evancz:


mgold:
Loving these new error-specific docs!

evancz:
Thanks :) Just have one more planned covering infinite types.

Issue #1143, opened 2015-11-02
[0.16 regression] Inferred a, annotated number causes error


mgold:
The following compiles in 0.15.1 but not in 0.16-alpha3.

```elm
px : number -> String
px x = toString px ++ "px"
```

The inferred type is `px : a -> String`, but it's reasonable for an annotation to make a type less general. And indeed, changing `a` to any specific type does not error. But `number`, `comparable`, and `appendable` all cause a type mismatch.

A good acceptance test is that `a` and `b` can both become a `number` (is it possible to specify that they don't need to be the same number?):

```elm
pointString : number -> number -> String    
pointString x y =    
  (px x) ++ (px y) 
```

laszlopandy:
It would be helpful if you can make a minimal self-contained test, add it to the [good examples folder](https://github.com/elm-lang/elm-compiler/tree/master/tests/test-files/good) and send a pull request. The PR will not be mergeable until the tests are fixed, but it will provide an easy way to check any fix and to make sure this regression doesn't come up again.

mgold:
I think the code blocks above are fine minimal examples, and can easily be added whenever a fix happens. Also, I assume under the Soundness directory?

evancz:
Just always include at http://sscce.org/ so I don't have to do copy/pasting, reading, blah blah blah, etc. It makes it easier if it is like "Here is the broken code" and "here is what I am seeing from the compiler with version X"

laszlopandy:
@mgold Yes Soundness directory sounds fine. But you should not wait until Evan has fixed it. You should create a PR, so Evan and check out the branch and easily run `cabal test` to verify it. Otherwise he has to spend time making sure he is fixing the right thing you asked for.

For example, there are also usually some issues about making an minimal example which will run in compile example tests (they have to be entirely self-contained without imports, which means `++` and `toString` are not available). If you make a PR where the tests fail with the right error, we can be sure he fixes the right thing.

mgold:
Okay, PR is out with a failing test.

jdudek:
I did some `git bisect` today and the first commit where the submitted file stopped compiling is https://github.com/elm-lang/elm-compiler/commit/a28acb21900535f3229e7ef924856b70298870c5. Not sure if it helps though, as that commit was a pretty big change :)

evancz:
Awesome, that is super helpful @jdudek, thank you for doing that! :D

PR #1157, opened 2015-11-11
Specific annotations tests

mgold:
Add (failing) tests for #1143.

In general, it should be possible to restrict a type by writing an annotation more specific than the inferred type. One test catches a 0.16-alpha regression in which doing so for `number` (also `comparable`, `appendable`) was disallowed. The other test verifies that doing so for concrete type `Int` works, which it always has, but it seemed like a good thing to test, and it allowed me to write the expected JS (which is identical for both tests).


mgold:
Merged with #1178. Thanks!

Issue #1146, opened 2015-11-03
No kind checking


mgold:
It is possible to use higher-kinded types in union types and type aliases:

```elm
type Impossible
    = A List
    | B (List Int Int)

type alias Uninhabited =
    { field : Dict String
    }
```

None of these types are possible, so while it's not hurting anyone, it's also kind of weird to let meaningless things be written down.

**Note:** Comment edited to summarize discussion and refinement in the further comments.

mgold:
Failing test submitted in PR.

jvoigtlaender:
Also, this
```elm
type A = B Maybe
```
is accepted, even though it is meaningless.

So the issue here is not really just about type aliasing. It is about a lack of kind checking.

Maybe rename the issue?

Also, [here](https://groups.google.com/d/msg/elm-discuss/GkNP-EbiHlI/obTnsas0BAAJ)'s a case where this (possibility of defining meaningless types this way) hit a user.

evancz:
It's not 100% obvious to me that higher-kinded type aliases should not be permitted, but it does seem reasonable to demand that a certain things have kind `Type`, such as record fields or values in a union type. Stuff that needs to be a value that can exist!

Heimdell:
Maybe, the compiler should perform a kind check at the declaration site? Is the kind of type atoms stored anywhere?

mgold:
Another kind issue: extending record aliases with non-aliases crashes the compiler.

```elm
type alias Foo a =
  { a | bar : Int }
  
type alias Bar = Foo (Int -> Int)

bar : Bar
bar = {bar = 2}
```
The dumb solution is to improve the error message found [here](https://github.com/elm-lang/elm-compiler/blob/967c8959674ba4d9d64e139e9d1c623973ac27d7/src/Type/Type.hs#L332-L341).

evancz:
Centralizing into #1373, will track from there.

evancz:
The development branch resolves this by requiring that things line up syntactically. It is not doing a kind check during type inference though. So in this case, it is fixed with:

```
-- TOO FEW ARGS ------------------------------------------------------- temp.elm

The `Dict` type needs 2 arguments, but I only see 1:

15|     { field : Dict String
                  ^^^^^^^^^^^
What is missing? Is there some typo?
```

Issue #1110, opened 2015-10-16
Type checker should forbid unbound type variables in type aliases


psfblair:
Example case:

    type MyData = MyData String
    type alias MyDataAlias a = MyData

Complete example here: http://share-elm.com/sprout/56213fb6e4b0966c193dd8ee

The type checker just ignores the type variable, but this should probably be an error instead -- either the author intended a phantom type, which can't be introduced by a type alias, or intended to bind the type variable, in which case the code is erroneous. I suggest having the type checker say something like: "You have an unbound type variable in a type alias. Phantom types are not supported through aliases. Maybe you meant ...?"

evancz:
Thanks for the report!

PR #1158, opened 2015-11-11
Test invalid type aliases

mgold:
This is a separate PR to add more failing tests for other issues, related to type aliases.

PhantomTypeAliases is a regression test for #1110. It passes in master.

MissingTypeVariables is a currently failing test for #1146. I have bundled them together to show that fixing one has not fixed the other.

### DIFF
--- a/tests/test-files/bad/MissingTypeVariables.elm
+++ b/tests/test-files/bad/MissingTypeVariables.elm
@@ -1,0 +1,1 @@
+type alias Bad = List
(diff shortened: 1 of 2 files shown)


evancz:
@laszlopandy had me activate a thing where we can't merge anything into master unless the tests pass. So I guess it makes sense to split this up?

mgold:
I don't think merging the passing tests is a high priority. It was more meant to demonstrate that the bug persists despite a similar bug being fixed.

I think it would be reasonable to merge my branch into the dev branch for this bug (whenever you get around to it) and merge tests and bugfix together.

evancz:
I think these are examples of the kind checking stuff. I'd like to just track it all in that issue.

PR #1159, opened 2015-11-11
Allow Char values to be passed through ports

jdudek:
Hi,

I’ve noticed that you can’t pass `Char` values through inbound and outbound ports. I don’t know if there is a good reason for that, so I thought I’d try to add support for passing chars.

The way it works is as follows:

* for inbound ports, a value given from JS is wrapped with `Utils.chr`
* for outbound ports the value is just passed as is—as a `String` object tagged with `isChar`. This may be not ideal, as the resulting object won’t be an `instanceof String`. So maybe it would be better to call `toString` on it.

In the inbound case, there is no check for whether only a single-char string was passed. Do you think we should check it too? I suppose it should be easy to add.

I also noticed that we’re fairly permissive when it comes to number values. For instance, you can pass a non-integer number to a port of type `Int` and no error is raised. I’ll be happy to fix that, I just wasn’t sure if it’s intentional or not.

I’ll appreciate any feedback on the PR :) Thanks!

### DIFF
--- a/src/AST/Variable.hs
+++ b/src/AST/Variable.hs
@@ -140,7 +140,7 @@ isPrimitive :: Canonical -> Bool
 isPrimitive v =
     case v of
       Canonical BuiltIn name ->
-          name `elem` ["Int","Float","String","Bool"]
+          name `elem` ["Int","Float","String","Char","Bool"]
       _ ->
           False
 
--- a/src/Generate/JavaScript/Port.hs
+++ b/src/Generate/JavaScript/Port.hs
@@ -134,6 +134,9 @@ toType tipe x =
           where
             from checks = check x checks x
 
+      T.Type (Var.Canonical Var.BuiltIn "Char") ->
+            check x JSString (obj ["_U","chr"] `call` [x])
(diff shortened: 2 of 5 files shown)


mgold:
Please open another issue for passing non-integers to Int ports, with a minimal reproduction.

As for Char going through ports, yes it ought to be possible, but for prioritization's sake, what's the use case? Can't you just pass strings and convert to Char in Elm?

jdudek:
Thanks for the reply.

Sure, I can convert strings, but it requires writing a converter—whereas for other simple data types I don’t have to. In case of a nested structure it’s just a bit annoying—I need to unwrap a structure 2-levels deep, just to map one datatype within it, and then wrap it again.

My use case is a letter game (let’s say it’s similar to Scrabble) and it was a natural choice to model letter tiles with `Char` datatype.

I’ll create the other issue for `Int`.

evancz:
I think it's better to track this as an issue. Lots of things are wanted for ports, so it'd be best to batch all those decisions and changes together.

Issue #1162, opened 2015-11-12
Ports of type Int accept non-integer numbers


jdudek:
Here’s the problem I mentioned in #1159.

See the following Elm program:

```elm
module Sample where

import Graphics.Element exposing (show)

main =
  let
    impossible =
      3 < int && int < 4

    msg =
      "Given integer: " ++ (toString int) ++
      (if impossible then " is bigger than 3 and smaller than 4" else "") ++
      ". Multiplied by 2: " ++ (toString (int * 2)) ++ "."
  in
    show msg

port int : Int
```

When run like this:

```javascript
Elm.fullscreen(Elm.Sample, { int: 3.14 })
```

It prints:

```
Given integer: 3.14 is bigger than 3 and smaller than 4. Multiplied by 2: 6.28.
```

(a full runnable example is here: https://gist.github.com/jdudek/d89c52f12b10f732d912)

I don’t know what would be the best expected behavior here. Either we convert values given to the port (e.g. with `parseInt`) or we raise an exception. We can check if a number is integer with [Number.isInteger](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) (we can have a function in `Utils` that falls back to the polyfill).

I can work on this issue, if you think it’s worth fixing :)

mgold:
JS -> Elm ports have a concept of invalid values, and I think we should build on that rather than silently converting floats to ints.

evancz:
Should be relatively simple to add in. Can this be a PR?

jdudek:
The PR is here: https://github.com/elm-lang/elm-compiler/pull/1164 :)

evancz:
Ah, thanks! Should make it into 0.16 then :)

evancz:
Alright, your fix just got merged. Thanks again! :D

PR #1164, opened 2015-11-13
Disallow non-integer numbers to be passed to Int ports

jdudek:
This should solve the issue I reported in #1162.

It adds additional checks performed when accepting a value to an incoming port of type `Int`.

The check is based on a polyfill for [Number.isInteger](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger). NaN and infinite values are handled correctly (and not accepted.

The output for the [test case](https://gist.github.com/jdudek/d89c52f12b10f732d912) I posted looks as follows:

```
Port Error:

    Regarding the port named 'int' with type:
    
        Int
    
    You just sent the value:
    
        3.14
    
    but it cannot be converted to the necessary type.
    Runtime error when sending values through a port.
    
    Expecting an integer but was given 3.14

Open the developer console for more details.
```

### DIFF
--- a/src/Generate/JavaScript/Port.hs
+++ b/src/Generate/JavaScript/Port.hs
@@ -27,6 +27,7 @@ task name expr portType =
 
 data JSType
     = JSNumber
+    | JSIntNumber
     | JSBoolean
     | JSString
     | JSArray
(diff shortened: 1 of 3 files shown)


laszlopandy:
:+1:

evancz:
Okay, needed to rebase, then cleaned up some stuff. New thing is at #1175!

jdudek:
Thank you!

PR #1165, opened 2015-11-15
List [a..b] syntax accepts floats

mgold:
The `[a..b]` list literal syntax expects `a` and `b` to be numbers, not ints. This leads to odd behavior when it's passed floats.

```
> [1.1..3.14]
[0.14000000000000012,1.1400000000000001,2.14,3.14]
> [-1.1 .. 2.3]
[-1.7000000000000002,-0.7000000000000002,0.2999999999999998,1.2999999999999998,2.3]
```
If someone else can think of a way in which this syntax makes sense, then I'm open to that, but otherwise I think it just needs to be removed. That is, the snippets above would be rejected with a int/float type mismatch error. Therefore this PR includes a failing "bad" test.

This behavior has been present since at least 0.12.3 (the latest version on share-elm).


laszlopandy:
I think the whole .. syntax should be removed in favour of:
List.intRange : Int -> Int -> List Int

Then we can have other function and all will be more explicit:
List.intRangeWithStep : Int -> Int -> List Int
List.floatRange : Float -> Float -> List Float

mgold:
That's not a bad idea, but it's certainly a bigger change than fixing a bug.

evancz:
Rather track this as an issue

PR #1167, opened 2015-11-16
Allow qualified names in case patterns

### LINKED ISSUE #951: Qualified names can not be used as type constructor arguments in `case` statements.
Bar.elm:

```elm
module Bar where

type Action = Baz | Qux
```

Foo.elm:

```elm
module Foo where

import Bar exposing (Action(Qux))

type Action = NoOp | BarAction Bar.Action

foo : Action -> Int
foo action =
  case action of
    NoOp -> 0
    BarAction Bar.Baz -> 1 -- ERROR!
    BarAction Qux -> 2 -- works

faa : Bar.Action -> Int
faa action =
  case action of
    Bar.Baz -> 0 -- works
    Bar.Qux -> 1 -- works
```

Compilation result:

    Error in Foo.elm:

    Parse error at (line 11, column 18):
    unexpected "."
    expecting whitespace, a newline, pattern, cons operator '::' or arrow (->)

Expected behaviour:
`BarAction Bar.Baz` should match without error.

sgillis:
This should fix #951. I didn't add any tests because it would require having multiple modules in the tests which doesn't make sense in this repo as [discussed in the elm-dev group](https://groups.google.com/forum/#!topic/elm-dev/-3_G-ixDLWM).

### DIFF
--- a/src/Parse/Pattern.hs
+++ b/src/Parse/Pattern.hs
@@ -98,7 +98,8 @@ patternConstructor =
         case v of
           "True"  -> return $ P.Literal (L.Boolean True)
           "False" -> return $ P.Literal (L.Boolean False)
-          _       -> P.Data (Var.Raw v) <$> spacePrefix term
+          _       -> P.Data (Var.Raw v) <$>
+                       spacePrefix (patternConstructor <|> term)
 
 
 expr :: IParser P.RawPattern


avh4:
Could you add a test that refers to something in core?

```elm
foo : Action -> Int
foo action =
  case action of
    NoOp -> 0
    BarAction Maybe.Nothing -> 1 -- ERROR!
    BarAction (Maybe.Just Maybe.Nothing) -> 2 -- ERROR!
    BarAction (Maybe.Just _) -> 3
```

mgold:
@avh4 No, core is not in scope during testing (as currently situated).

evancz:
Thanks for the PR! I ran into this at an Elm meetup recently. I think this PR not handle parentheses properly. I'll try to fix it another way.

PR #1170, opened 2015-11-17
Correctly parse Foo.Bar.record.field

### LINKED ISSUE #520: cannot access field of qualified record (Graphics.Collage.defaultLine.color)
Syntax like `Graphics.Collage.defaultLine.color` won't compile; you instead have to write `.color Graphics.Collage.defaultLine`, which is clunky. Since modules are always upper-case and variables are always lower-case (and data constructors don't have fields), I don't think there's any ambiguity. Has this been discussed before?

sgillis:
Allow access to fields of qualified records

### DIFF
--- a/src/Parse/Helpers.hs
+++ b/src/Parse/Helpers.hs
@@ -295,8 +295,8 @@ accessible exprParser =
                 end <- getMyPosition
                 return . A.at start end $
                     case rootExpr of
-                      E.Var (Variable.Raw name@(c:_))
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
There seem to be merge conflicts here.

evancz:
Changing the parser can be tricky, so I'd like to do it myself. I'll track it from the original issue. Thanks for the PR though!

PR #1173, opened 2015-11-17
Fix #1172, destruct chars before pattern matching

### LINKED ISSUE #1172: Pattern matches on Chars fail
The following code goes to the wildcard branch.

```elm
import Html

main =
  Html.text <|
    case 'A' of
      'A' ->
        "hello"

      _ ->
        "letter A was not seen"
```

It is because of the internal representation of `Char`. I'll have a fix in a few minutes.

evancz:


laszlopandy:
This is a good case to have an expected output test.

laszlopandy:
Here's a PR for the test: https://github.com/elm-lang/elm-compiler/pull/1177

PR #1180, opened 2015-11-18
Stop using ↓ because of dumb British windows

evancz:
Seems like unicode makes it go crazy pretty reliably


mgold:
Such a pity, would be nice to have the compiler intelligently detect if unicode is supported, but I agree this is the right thing to do to get 0.16 out the door.

PR #1182, opened 2015-11-18
Allow upper case names for now

evancz:
Instead it makes sense to make package.elm-lang.org disallow publishing
of packages with capital letters. This way things keep working for 0.15
people.

### DIFF
--- a/src/Elm/Package.hs
+++ b/src/Elm/Package.hs
@@ -74,9 +74,6 @@ validate str =
   else if elem '_' str then
       Left "Underscores are not allowed in package names."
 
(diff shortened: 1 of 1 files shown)


laszlopandy:
Instead I would let no new packages use upper case, and whitelist the current set of uppercase packages until we can figure out how to convert them.
The fastest way to implement this: hardcode all the package names from package.elm-lang.org which have uppercase letters. And check against the hardcoded list when publishing a package.

laszlopandy:
Here you go: #1184

PR #1183, opened 2015-11-18
Adjust `main` check to look for new Element type

evancz:


zgotsch:
Hey @evancz, I am seeing an error that seems like it could be related to this:

```
The 'main' value has an unsupported type.

52│ main : Signal Element
    ^^^^^^^^^^^^^^^^^^^^^
I need an Element, Html, (Signal Element), or (Signal Html) so I can render it
on screen, but you gave me:

    Signal Element

Detected errors in 1 module.
```

This is in the release version of 0.16. The code was working in one of the release candidates for 0.16, but broke when I upgraded. Is there a thread about this somewhere?

zgotsch:
Immediately after typing this I realized it was probably a problem with my libs. Running `elm-package install` didn't do anything, but blowing away elm-stuff and reinstalling packages fixed my problem.

PR #1184, opened 2015-11-18
Whitelist uppercase package names

laszlopandy:


evancz:
Doing a modified form of this in #1186

Great idea!

PR #1189, opened 2015-11-19
Friendlier missingBranchesMessage

henrik:
I think the "!" makes it read less friendly that it could be. This is also more consistent with other messages in this file.


evancz:
What did it sound like to you?

I guess I tend to use exclamation marks a lot, and it's more about having an excited tone than a yelly one.

Anyway, can you show the full error message that you got? It maybe made sense in code but the full error message looks weird.

henrik:
@evancz I think it was actually from reading the examples here: http://elm-lang.org/blog/compilers-as-assistants



Definitely not a big deal, just a thought :)

evancz:
I think I'd like to leave it for now?

henrik:
@evancz Sure, up to you :)

PR #1190, opened 2015-11-20
Fix ANSI issue everywhere

evancz:
cc @jvoigtlaender

I thought that displayIO was for ANSI and displayS was for not-ANSI. It
turns out that displayS will add the ANSI codes into the string if you
are on not-Windows. This meant that garbage characters were ending up
in the output of JSON errors or reactor errors

@rtfeldman, I’ll try to get a new set of binaries ready for linux
32-bit and mac

### DIFF
--- a/src/Elm/Utils.hs
+++ b/src/Elm/Utils.hs
@@ -59,7 +59,9 @@ run :: (MonadError String m, MonadIO m) => String -> [String] -> m String
 run command args =
   do  result <- liftIO (unwrappedRun command args)
       case result of
-        Right out -> return out
+        Right out ->
+          return out
+
         Left err ->
           throwError (context (message err))
   where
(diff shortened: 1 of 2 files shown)


rtfeldman:
:+1:

rtfeldman:
@evancz I assume you mean Windows and 32-bit Linux? Travis has the other binaries covered.

evancz:
I believe Windows is fine because of the particulars of `displayS` where on Windows it already drops things, so I wasn't going to bother.

evancz:
I want to figure out https://github.com/elm-lang/elm-reactor/issues/156 as well though. I am struggling at the moment though :/

evancz:
What do you think about the plan to let it be broken until the new reactor is ready?

Issue #1200, opened 2015-11-24
Compiler generates invalid JavaScript instead of rejecting syntactically invalid Elm (qualified tags in union types)


jvoigtlaender:
From this Elm code:

``` elm
module Test where

type A = B.C
```

the compiler generates this JavaScript code:

``` javascript
Elm.Test = Elm.Test || {};
Elm.Test.make = function (_elm) {
   "use strict";
   _elm.Test = _elm.Test || {};
   if (_elm.Test.values) return _elm.Test.values;
   var _U = Elm.Native.Utils.make(_elm),
   $Basics = Elm.Basics.make(_elm),
   $Debug = Elm.Debug.make(_elm),
   $List = Elm.List.make(_elm),
   $Maybe = Elm.Maybe.make(_elm),
   $Result = Elm.Result.make(_elm),
   $Signal = Elm.Signal.make(_elm);
   var _op = {};
   var B.C = {ctor: "B.C"};
   return _elm.Test.values = {_op: _op,B.C: B.C};
};
```

which fails at runtime like so:

```
   var B.C = {ctor: "B.C"};
        ^
SyntaxError: Unexpected token .
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:906:3
```

laszlopandy:
Please send a PR with the example in the good/ directory. For example, like this PR:
https://github.com/elm-lang/elm-compiler/pull/1179/files
Then when Evan fixes the issue, he will have a easy test case to check if the right output is produced.

jvoigtlaender:
I'm afraid your request makes no sense. There is no "correct code" here. The compiler should reject the Elm program!

deadfoxygrandpa:
@jvoigtlaender there's also a "bad" directory for files that should not compile. I added a case for this issue in the above PR.

jvoigtlaender:
Okay, that makes more sense. :smile:

agrafix:
I can confirm this bug, ran into it, too. ( https://github.com/agrafix/elm-html-shorthand/commit/a674950a60ffce48d0ac85f4c9a40e4482c94c1f )

ccapndave:
And me too :)

jfeaver:
[...]

PR #1201, opened 2015-11-24
Add bad case for bug in type definitions

deadfoxygrandpa:
Addresses issue #1200, this file should not compile.


mgold:
Wow, this actually does fail - I assumed it would have errored with "B is not in scope". When verifying a potential fix, you have to make sure it fails for the right reason!

deadfoxygrandpa:
Are you sure? I think it fails because it's not supposed to compile, but it does compile. At least, that's how I'm reading this:
```
  Bad Tests:

    tests/test-files/bad/DotInType.elm: [Failed]

Compilation succeeded but should have failed

    tests/test-files/bad/IncompletePatternMatch.elm: [OK]

    tests/test-files/bad/InfiniteType.elm: [OK]

    tests/test-files/bad/NonElementMain.elm: [OK]

    tests/test-files/bad/Strings/ExtraClose.elm: [OK]
```
Maybe the output is confusing, because it looks like it's saying all of the "bad" tests except the new one I just added actually compile. But all of these tests were correct in all the other recent builds, I think?

mgold:
> I think it fails because it's not supposed to compile, but it does compile.

Exactly. I typed the test into the REPL and got a JS error, so it compiled (but it shouldn't have). My concern is that I expected the compiler to reject the program because B was not defined, but if we imported B, the bug could still exist.

ashnur:
I found this error while tried to define a type like this:
```
type KeyCode = Char.KeyCode
```
because I had some issues after updating to 0.16, and I was also very tired so I did the worst thing, just guessed something, then it compiled (before there was a compiler error regarding KeyCode not being present). Then I tried to look at it in the browser and it failed.

jvoigtlaender:
Can this test case be merged, to hopefully facilitate fixing bug https://github.com/elm-lang/elm-compiler/issues/1200 sooner?

evancz:
I'd rather track this as an issue. It's hard to manage PRs when there are a bunch of ones in there that just need to sit for a while.

PR #1215, opened 2015-12-01
Update changelog for 0.16

igrep:
https://github.com/elm-lang/elm-compiler/issues/1191#issuecomment-160890566

PR #1217, opened 2015-12-01
Grammatical correction of #1215

igrep:
Sorry, I should've found before sending https://github.com/elm-lang/elm-compiler/pull/1215 :bow:
I forgot to append period and breakline.


evancz:
I have some changes in #1216 that also resolve this. And don't worry about it, I'm very grateful for your help! :D

PR #1220, opened 2015-12-03
Tests for #951: Allow qualified names in case patterns

jdudek:
I wrote a test for the pull request (#1167) submitted by @sgillis.

I’m not sure if running compiler to get the interface is the best solution… I tried the alternative, i.e. with awfully lot of effort I was able to construct the interface as AST. It looked more or less like this: https://gist.github.com/jdudek/1b6d0a17c80ca791563d I felt it was less readable, so I submitted this version.

I also wasn’t sure if cheating by incomplete pattern match for compilation result is OK… perhaps there is a better way.

Anyway, I hope it will be useful :)


evancz:
I ended up fixing it a different way. I don't think the tests can handle imports, but I'm not certain.

PR #1221, opened 2015-12-03
Tests for #520: cannot access field of qualified record 

### LINKED ISSUE #520: cannot access field of qualified record (Graphics.Collage.defaultLine.color)

jdudek:
Here’s another test—for PR #1170 (which fixes #520).

### DIFF
--- a/src/Parse/Helpers.hs
+++ b/src/Parse/Helpers.hs
@@ -295,8 +295,8 @@ accessible exprParser =
                 end <- getMyPosition
                 return . A.at start end $
                     case rootExpr of
-                      E.Var (Variable.Raw name@(c:_))
-                        | isUpper c ->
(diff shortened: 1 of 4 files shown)


AppSynergy:
Would love for this to go in!

jvoigtlaender:
There seem to be merge conflicts here.

jdudek:
Thanks! I’ll try to merge it with current master today evening UTC.

evancz:
Will add tests when the underlying fix is made. Rather not do it through PR until then.

PR #1249, opened 2016-01-06
Add tests for the parser

avh4:
I wrote this suite of tests while working on elm-format, and thought it would be good to port them over to elm-compiler.  This will allow refactoring of the parser to be done more safely, and will also let me track any subtle syntax changes that happen between versions so that elm-format can be properly updated.

### DIFF
--- a/src/AST/Expression/General.hs
+++ b/src/AST/Expression/General.hs
@@ -54,6 +54,7 @@ data Expr' ann def var typ
     -- for type checking and code gen only
     | Port (PortImpl (Expr ann def var typ) typ)
     | GLShader String String Literal.GLShaderTipe
+    deriving (Eq, Show)
 
 
 -- PORTS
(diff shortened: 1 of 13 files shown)


avh4:
@evancz thoughts on whether you think this is worthwhile?

rtfeldman:
Seems pretty useful right about now, given the recent issues filed about regressions! 😄

cc @evancz

mgold:
Yes! And all of the new parser regressions should be added to the test suite if it doesn't catch them already.

avh4:
Actually, I think this is not what we want to do.  These unit tests that expect on the parsed AST have been extremely brittle and painful to maintain in elm-format.  I think this PR should be dumped, and instead there should be more tests of the type that already exist: having example Elm files, and simply verify that they compile (and/or verify that they produce a specific js output).

Issue #1223, opened 2015-12-07
shadowed variable can lead to runtime error in generated JS


challiwill:
We have this function:

``` elm
viewBuildDuration : Time.Time -> BuildDuration -> Html
viewBuildDuration now duration =
  Html.dl [class "build-times"] <|
    case (duration.startedAt, duration.finishedAt) of
      (Nothing, _) ->
        []

      (Just startedAt, Nothing) ->
        labeledRelativeDate "started" now startedAt

      (Just startedAt, Just finishedAt) ->
        let
          duration =
            Duration.between (Date.toTime startedAt) (Date.toTime finishedAt)
        in
          labeledRelativeDate "started" now startedAt ++
            labeledRelativeDate "finished" now finishedAt ++
            labeledDuration "duration" duration
```

Which compiles to this:

``` javascript
   var viewBuildDuration = F2(function (now,duration) {
      return A2($Html.dl,
      _U.list([$Html$Attributes.$class("build-times")]),
      function () {
         var _p0 = {ctor: "_Tuple2",_0: duration.startedAt,_1: duration.finishedAt};
         if (_p0._0.ctor === "Nothing") {
               return _U.list([]);
            } else {
               if (_p0._1.ctor === "Nothing") {
                     return A3(labeledRelativeDate,"started",now,_p0._0._0);
                  } else {
                     var _p2 = _p0._0._0;
                     var _p1 = _p0._1._0;
                     var duration = A2($Duration.between,$Date.toTime(_p2),$Date.toTime(_p1));
                     return A2($Basics._op["++"],
                     A3(labeledRelativeDate,"started",now,_p2),
                     A2($Basics._op["++"],A3(labeledRelativeDate,"finished",now,_p1),A2(labeledDuration,"duration",duration)));
                  }
            }
      }());
   });
```

This doesn't work because `duration` in the nested function needs to be named differently. Currently it leads to a runtime error: "Cannot read property 'startedAt' of undefined".

jvoigtlaender:
Do you really have the line
```elm
          var duration = A2($Duration.between,$Date.toTime(_p2),$Date.toTime(_p1));
```
in your *Elm* code? It looks like that above. Maybe clean up the code examples?

challiwill:
Oops, copy and paste mistake. Thanks for catching it.

I've removed it from the above example.

gyzerok:
+1 for fixing this :)

jvoigtlaender:
[...]

PR #1253, opened 2016-01-07
Add test case for shadowing bug

jvoigtlaender:
This is a test case for https://github.com/elm-lang/elm-compiler/issues/1223.

The test case failing at the moment is of course expected. Hopefully, it will turn green when bug https://github.com/elm-lang/elm-compiler/issues/1223 is fixed.


jvoigtlaender:
The test case failing at the moment is of course expected. Hopefully, it will turn green when bug https://github.com/elm-lang/elm-compiler/issues/1223 is fixed.

jvoigtlaender:
The point here being that the current compiler / code generator outputs the following:
```js
var _elm_lang$core$Main$bad = function (arg) {
	return function (x) {
		return x;
	}(
		function () {
			var _p0 = arg;
			if (_p0.ctor === 'Just') {
				return true;
			} else {
				var arg = true;
				return arg ? false : arg;
			}
		}());
};
```
where it should instead output something like the following:
```js
var _elm_lang$core$Main$bad = function (arg) {
	return function (x) {
		return x;
	}(
		function () {
			var _p0 = arg;
			if (_p0.ctor === 'Just') {
				return true;
			} else {
				var _p1 = true;
				return _p1 ? false : _p1;
			}
		}());
};
```

evancz:
Let's do the tests after the fix. Do not want a bunch of open PRs sitting around. Hard to see what is going on.

PR #1256, opened 2016-01-09
Add custom error message if tab character is found in parsing.

laszlopandy:
Given this file:
```elm
list =
<tab>[
<tab><tab>1
<tab>]
```

Elm will now give this error message:
```
-- SYNTAX PROBLEM -------------------------- TabIndent.elm

A tab character was found, but all whitespace (including indentation) must be
spaces not tabs.

3│ 	[
   ^
I am looking for one of the following things:

    whitespace

```

### DIFF
--- a/src/Parse/Helpers.hs
+++ b/src/Parse/Helpers.hs
@@ -346,7 +346,7 @@ badWhiteSpaceMessage =
 
 failIfTabFound :: IParser ()
 failIfTabFound =
-  do  foundTab <- option False (lookAhead (char '\t') >> return True)
+  do  foundTab <- option False (lookAhead (char '\t') >> return True) <?> Syntax.tab
       if foundTab
         then fail tabsErrorMessage
         else return ()
--- a/src/Parse/Helpers.hs
+++ b/src/Parse/Helpers.hs
@@ -348,13 +348,8 @@ failIfTabFound :: IParser ()
 failIfTabFound =
   do  foundTab <- option False (lookAhead (char '\t') >> return True) <?> Syntax.tab
       if foundTab
-        then fail tabsErrorMessage
+        then fail Syntax.tab
         else return ()
-
-
-tabsErrorMessage :: String
-tabsErrorMessage =
-  "A tab character was found, but all whitespace (including indentation) must be spaces not tabs."
 
 
 -- Just eats whitespace until the next meaningful character.
--- a/src/Parse/Helpers.hs
+++ b/src/Parse/Helpers.hs
@@ -327,7 +327,9 @@ spaces =
 
 forcedWS :: IParser String
 forcedWS =
-  do  ws <- many1 (spaces <|> newline)
+  do  failIfTabFound
+      ws <- many1 (spaces <|> newline)
+      failIfTabFound
       column <- sourceColumn <$> getPosition
       if column == 1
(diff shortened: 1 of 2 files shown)


mgold:
LGTM.

laszlopandy:
@evancz

src/Parse/Helpers.hs
```diff
@@ -327,7 +327,9 @@ spaces =
 
 forcedWS :: IParser String
 forcedWS =
-  do  ws <- many1 (spaces <|> newline)
+  do  failIfTabFound
+      ws <- many1 (spaces <|> newline)
+      failIfTabFound
```
evancz: Could these three lines be condensed into the following?

```haskell
ws <- many1 (spaces <|> newline <|> failOnTab)
```

src/Parse/Helpers.hs
```diff
@@ -342,6 +344,19 @@ badWhiteSpaceMessage =
   ++ "some spaces here:"
 
 
+failIfTabFound :: IParser ()
+failIfTabFound =
+  do  foundTab <- option False (lookAhead (char '\t') >> return True)
+      if foundTab
+        then fail tabsErrorMessage
+        else return ()
```
evancz: Along with the other change, could this be reduced to the following?

```haskell
failOnTab :: IParser a
failOnTab =
  do  char '\t'
      fail tabsErrorMessage
```

evancz:
In addition to those two code things, I feel like it is maybe weird that it says it is expecting whitespace in the new error. I forget exactly how [this code](https://github.com/elm-lang/elm-compiler/blob/master/src/Reporting/Error/Syntax.hs#L247-L299) works, but would you mind reading through it til you see the tricks so you can assess if we can make that part better too?

src/Parse/Helpers.hs
laszlopandy: Yes, this is what I tried initially:
```haskell
ws <- many1 (spaces <|> newline <|> (char '\t' >> fail "tab error"))
```

The problem is that it consumes the tab character before failing so the error message position is way out of whack:
```
-- SYNTAX PROBLEM ------------------------------- tests/test-files/good/Tabs.elm

tab error

3│   	[
           ^
Maybe <http://elm-lang.org/docs/syntax> can help you figure it out.
```

To fix that I tried adding `lookAhead`, but that makes it worse. `many1` works in a way that parsers have to fail to indicate that it should stop repeating. If a parser inside `many1` fails *without consuming input* then `many1` returns and let's the next parser take over. 

In short, you cannot put a `lookAhead` inside a `many1` and still propagate an error. This is why `failIfTabFound` has to be outside it.

laszlopandy:
I've added two more commits to improve the error messages. The first removes "\t" from the list of expected things if whitespace is expected.

The second commit changes the error message when we encounter a tab:
```
-- SYNTAX PROBLEM ------------------------------- tests/test-files/good/Tabs.elm

A tab character was found, but all whitespace (including indentation) must be
spaces not tabs.

3│   	[
     ^
I am looking for spaces, not tabs.
```

src/Parse/Helpers.hs
evancz: Ah, interesting. Makes sense!

evancz:
Awesome, excited to have this in the next release!

PR #1257, opened 2016-01-09
Allow trailing commas in lists and records.

laszlopandy:
The only concern I have is that `spaceySepByTrailing1` is recursively defined. I don't know if that has any impact on the performance of parsec. But I could not figure out any other way to implement optional trailing commas.

I considered allowing optional trailing commas in tuples too but I don't know if it is useful. It requires more work because we need to explicitly reject the parsing of  `(1,)` because it would be interpreted as an expression in parens (single element tuples don't exist in Elm).

@avh4
@rtfeldman


laszlopandy:
@evancz

src/Parse/Helpers.hs
```diff
@@ -156,6 +156,19 @@ spaceySepBy1 sep parser =
...
 
+spaceySepByTrailing1 :: IParser b -> IParser a -> IParser [a]
+spaceySepByTrailing1 sep parser =
+  do  value <- parser
+      choice
+        [ do  try (padded sep)
+              choice
+                [ (value:) <$> spaceySepByTrailing1 sep parser
+                , return [value]
+                ]
```
avh4: The `choice` can be simplified with `option`,  something like `option [value] $ (value:) <$> spaceySepByTrailing1 sep parser`

avh4:
We should see if we can completely get rid of `commaSep` and `commaSep1`--anywhere those are use we should consider whether we can allow trailing spaces there as well.

evancz:
Easy enough for me to change the parser, I'd rather add this myself at the appropriate time.

PR #1265, opened 2016-01-20
Fix unsafe list patterns in error reporting.

srijs:
Hi.

This fixes up some unsafe pattern matching by using non empty lists where appropriate.

Please let me know if this is welcome.

Cheers!


evancz:
Thanks! I'd rather not add more dependencies though. I am willing to tolerate the risk for now.

srijs:
We could have done the same thing with tuples of (a, [a]), but sure, I'll leave that up to you ;)

PR #1290, opened 2016-02-16
Avoid repeated arguments at compile time. Fix #809

### LINKED ISSUE #809: Detect repeat argument names at compile time
The following code

```elm
sum x x = x + x
mySum = sum 1 2
```

Generates the error

```
Uncaught SyntaxError: Strict mode function may not have duplicate parameter names
```

Which is good, except that the error message appears only in the console. Instead, the webpage only shows
```
Cannot read property 'make' of undefined
    Open the developer console for more details.
```

Which makes debugging really confusing.

yblein:
A small patch to get started with the Elm codebase.
It allows to detect repeated argument names at compile time and not at runtime.
It also prevents repeated arguments nested in patterns like
```
sum x (x, y) = x + y
```
that where not crashing, yet meaningless.


mgold:
Whoa great work!

Currently checking for duplicates is O(n^2). It probably won't matter because argument lists are short, but you could do better, perhaps like so:

```haskell
import Data.List (sort)

firstDuplicate :: [String] -> Maybe String
firstDuplicate vars =
  let
    sorted = sort vars
    pairs = zip sorted (tail sorted)
    filtered = filter (uncurry (==)) pairs
  in
    case filtered of
      [] -> Nothing
      (x,_):_ -> Just x
```

src/Validate.hs
```diff
@@ -194,6 +194,38 @@ checkDefinition (Valid.Definition pattern body _) =
                 Result.throw (R.merge start end) (Error.BadFunctionName (length args))
 
 
+
+checkArguments :: [Pattern.Pattern R.Region var] -> Result.Result wrn Error.Error ()
```
yblein: I guess there a type synonym for "Pattern.Pattern R.Region var" but I couldn't find it.

yblein:
I also though that the O(n^2) in not an issue considering the arity of a function.
A quick benchmark shows that your solution is faster when the arity is greater than 5. I used a test list with no duplicates since it is the valid case. I have no idea about the average arity though.

mgold:
It's probably fine. We're talking milliseconds right?

evancz:
Thanks @yblein, I'm planning to go through a bunch of PRs in the lead up to 0.17. This one seems nice, so expect to hear from me again when 0.17 gets closer.

@mgold, I am slightly worried about "contributor fatigue" where folks without merge rights run contributors through a bunch of questions and concerns that do not necessarily help the review process. In the end, it's harder to figure out what the PR is, and the contributor feels like they did a bunch of extra work to satisfy "people" not knowing that it would have no direct correlation with merging or not. Like, if I come along later and have a similar question or different preference, the contributor is going to be like "blahhh, I already answered this, this is crazy!!! :( :( :("

In this case, it seems okay, but I have seen you and Janis get into pages of discussion on certain issues, and I'm not sure it's in the best interest of the contributor or their issue/pr. Anyway, just something to be aware of!

mgold:
No worries, I'll let you handle things. Thanks for asking nicely.

evancz:
Merged this into `dev` with a few modifications. I switched to a O(n * log n) implementation that takes only one pass over the list, so it should be the fastest version of things. Thanks @yblein!

PR #1299, opened 2016-02-29
Change 2015 to 2016

Gaafar:


evancz:
Thanks! Something is up with merging, so I added this in myself.

PR #1307, opened 2016-03-05
Add missing "the" in the error message for unbound type variables

leocassarani:
Change "type checker" to "the type checker".


evancz:
Something is up with merging, but I am adding this by hand. Thanks!

PR #1324, opened 2016-03-22
Don't allow duplicate fields in record types or aliases

### LINKED ISSUE #1323: Duplicate field definition
Hey,

It seems that the compiler allow me to define the same field twice, even with a __different type__, it just takes the latest definition but I would've expect it will raise an error.

Example:
```elm
type alias Model =
  { id: Int
  , a: Int
  , a: String
  }
```

mgold:
Fixes #1323 by disallowing duplicate fields in type aliases of records [EDIT: and in type annotations]. Includes ~~a test~~ two tests.

Evan, I realize this is something you could fix fairly quickly and in your preferred style. But this was an informative exercise for me, and I thought I'd at least give you the option of merging.

### DIFF
--- a/src/Parse/Type.hs
+++ b/src/Parse/Type.hs
@@ -1,7 +1,7 @@
 {-# OPTIONS_GHC -Wall -fno-warn-unused-do-bind #-}
 module Parse.Type where
 
-import Data.List (intercalate)
+import Data.List (intercalate, group, sort)
 import Text.Parsec ((<|>), (<?>), char, many, optionMaybe, string, try)
 
 import qualified AST.Type as Type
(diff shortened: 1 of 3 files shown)


jvoigtlaender:
@mgold, you trying to fix this in the wrong place in the compiler?

The following is also conceptually incorrect, but accepted by the current compiler version, and not touched at all by your fix:
```elm
f : { foo : String , foo : Int } -> Int
f r = r.foo
```
Certainly a proper fix of the underlying issue here should make sure this function (specifcally its annotation) is rejected as well.

mgold:
Valid point! This should be handled when types (whether annotations or aliases) are parsed.

I'm going to poke at the parser code and decide if I want to take a second try or close the issue.

mgold:
@jvoigtlaender I've added your code as a second test. Thanks again.

I've addressed the issue in the parsing code and it's a fairly small change. I can tell that the tests pass (are rejected), but I don't know enough about cabal etc to actually compile them manually and see what the error looks like, so there's that caveat.

mgold:
I managed to figure out how to run the patched compiler, and while it correctly detects the errors, the message is extremely unhelpful. I'm going to figure out a better way to do this and open a clean issue.

Issue #1214, opened 2015-12-01
share type variables between nested type annotations


evancz:
Right now you can write things like this:

```elm
id : a -> a
id x =
  let
    y : a
    y = x
  in
    y
```

This code is "obviously correct" but right now type variables are not shared between sub annotations. That means the `y : a` is talking about a fresh type variable, not the one that appears in `id : a -> a`. This is confusing.

Maybe it'd be good to figure out how to share them. Are there any problems that may come from this? I believe OCaml does it with no issues. Can anyone confirm or deny that?

jvoigtlaender:
I cannot speak about OCaml, but Haskell has this feature. You have to explicitly turn it on there: https://wiki.haskell.org/Scoped_type_variables.

jvoigtlaender:
Let me add that I don't think we want this to be a feature you have to explicitly turn on, like in Haskell. Instead, it should always be turned on. (It would be better if it were always turned on in Haskell.)

evancz:
Yeah, I have seen how it works in Haskell. I'm just wondering if there are any nice principled reasons for having the default behavior of Haskell. I don't know of any, and I'm like 99% certain OCaml does what you'd expect. I just feel like maybe there is a good reason that I don't know about.

mgold:
I'm in favor of this, since you cannot currently write some type annotations.

To capture the desired behavior: "let-bound type annotations have in scope type variables from the top-level definition and any parent let-expressions." Because these type variables are "in scope", they are not "fresh", and therefore should unify.

thSoft:
I was just bitten by this, getting `The type annotation does not match its definition. ... Hint: A type annotation is clashing with itself or with a sub-annotation.`, which hints in this direction, but does not make it obvious that the cause of the problem is this issue. (Then I remembered. :))

sgraf812:
I was bitten by the same error as @thSoft. In Haskell, the type variable would at least have a different name (e.g. `a1`). This may be confusing, too (after all, where would `a1` come from in your code?), so I'll suggest to emit a warning when there's a scoped type signature. Or just change to the `ScopedTypeSignatures` behavior.

evancz:
This works in the latest version of the `dev` branch.

Issue #1192, opened 2015-11-20
Wrong "expected" type from the annotation


Janiczek:
The "expected" and "actual" rows are the same when they probably shouldn't be.

```elm
The type annotation for `shrinker` does not match its definition.

17│ shrinker : Shrinker (List (Maybe MyType))
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
The type annotation is saying:

    Shrinker (Maybe MyType)

But I am inferring that the definition has this type:

    Shrinker (Maybe MyType)

```

I would have expected it to say:

```diff
-The type annotation is saying:
-
-    Shrinker (Maybe MyType)
+The type annotation is saying:
+
+    Shrinker (List (Maybe MyType))
```

The minimal code:

```elm
module Main (..) where

import Lazy.List exposing (empty)
import Shrink exposing (Shrinker, maybe)


type MyType
    = A
    | B


shrinkerMyType : Shrinker MyType
shrinkerMyType action =
    empty


shrinker : Shrinker (List (Maybe MyType))
shrinker =
    maybe shrinkerMyType
```

The relevant dependencies:

```
    "dependencies": {
        "NoRedInk/elm-lazy-list": "2.0.0 <= v < 3.0.0",
        "NoRedInk/elm-shrink": "1.0.3 <= v < 2.0.0",
        "elm-lang/core": "3.0.0 <= v < 4.0.0",
    },
    "elm-version": "0.16.0 <= v < 0.17.0"
```

jacob-tock:
I ran into another instance of what is probably the same bug, plus maybe another related bug. In Elm 0.16, the following code:

```elm
module Buggy where

import Debug

type alias Alias a = a -> a

convert : (a -> b) -> Alias b -> Alias a
convert _ = Debug.crash "unimplemented"

applyToSecond : Alias b -> Alias (a, b)
applyToSecond _ = Debug.crash "unimplemented"

wrongTypeFunction : Int -> String
wrongTypeFunction _ = Debug.crash "unimplemented"

badError : Alias value -> Alias Bool
badError alias =
  convert wrongTypeFunction (applyToSecond alias)
```

produces two error messages, both displaying the same type as both expected and actual:

```
-- TYPE MISMATCH ----------------------------------------------------- Buggy.elm

The type annotation for `badError` does not match its definition.

16│ badError : Alias value -> Alias Bool
               ^^^^^^^^^^^^^^^^^^^^^^^^^
The type annotation is saying:

[...]

PR #1342, opened 2016-04-13
Create issue template to try to reduce duplicates

mgold:
New issues will automatically be pre-populated with the contents of this file. This should hopefully cut down the volume of issues that we're seeing with the same bugs.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,11 @@
+Thank you for reporting an issue with Elm! Before submitting, please make sure it's not a commonly reported issue:
+
+* If you're getting a runtime error, make sure you don't have any recursively defined values (see #873).
+* If you're having trouble with type annotations in a let-expression, see #1214.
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
Add https://github.com/elm-lang/elm-compiler/issues/1214? It's also something that trips up people a lot, because they think that (type variables in) inner annotations mean something that they actually don't mean.

mgold:
Done.

jvoigtlaender:
@evancz, having this template added would probably help, a lot, the people who do the looking at new (and often redundant) issues reported in this repository almost daily. That potential good effect would be active immediately after addition of the template.

jvoigtlaender:
https://github.com/elm-lang/elm-compiler/issues/1192 is supposed to be fixed now.

mgold:
Done. Also these can be merged with GitHub's new squashing feature 🎉

mgold:
If this was ever to get merged, it should also mention the error message catalog.

evancz:
Nice, really like the text! Will merge this in by hand.

evancz:
Did some things in #1437

PR #1350, opened 2016-04-18
Disallow union types containing unused type variables

mgold:
The following is rightly valid Elm:
```elm
type Tree = Node | Leaf Int Tree Tree
```
The following is accepted by the 0.16 compiler but shouldn't be:
```elm
type Tree a = Node | Leaf Int Tree Tree
```
It has an unused type variable, and this should be rejected as it is for type aliases. The code to detect and report this is fairly mechanical, following the lead from aliases, and extracting out functions where appropriate.

I also improved the formatting of an error message and a related helper function, as described in the second commit message.

There's a simple test file that I added, and I was also able to get the patched compiler running to verify that the bad code is caught and has a nice message.

Rebased onto all the 0.17 changes so this should merge without conflicts.

### DIFF
--- a/src/Reporting/Error/Helpers.hs
+++ b/src/Reporting/Error/Helpers.hs
@@ -72,14 +72,17 @@ reflowParagraph paragraph =
 commaSep :: [String] -> String
 commaSep tokens =
   case tokens of
-    [token] ->
-      " " ++ token
+    [] ->
+      ""
 
-    [token1,token2] ->
-      " " ++ token1 ++ " and " ++ token2
+    [token] ->
+      token
+
+    [token1, token2] ->
+     token1 ++ " and " ++ token2
 
     _ ->
-      " " ++ List.intercalate ", " (init tokens) ++ ", and " ++ last tokens
+      List.intercalate ", " (init tokens) ++ ", and " ++ last tokens
 
 
 capitalize :: String -> String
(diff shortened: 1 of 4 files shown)


jvoigtlaender:
> It has an unused type variable, and this should be rejected as it is for type aliases.

Why should this be rejected? There are uses of this feature one might see as legitimate. When unused variables were outlawed for type aliases (https://github.com/elm-lang/elm-compiler/issues/1110), nothing was done about union types. Possibly on purpose.

mgold:
> There are uses of this feature for union types that one might see as legitimate.

Such as?

jvoigtlaender:
For one thing, a lot of `core` and other important packages will immediately fail to compile if your change is merged, because of lines like [this](https://github.com/elm-lang/core/blob/master/src/Task.elm#L44).

But more importantly, various applications of so-called phantom types. See https://wiki.haskell.org/Phantom_type, for example(s), but there is a multitude of others as well. It's true that not every type magic that Haskell offers needs to be present in Elm. But this does not automatically mean that everything that Haskell offers as type level modelling should be looked badly upon. There are lightweight uses of phantom types (e.g., tracking physical units or unit systems; or something like the validated/unvalidated example from the above wiki page) that it would be a pity to miss out on.

mgold:
Fair enough.

PR #1352, opened 2016-04-19
Validate record types to have no duplicate fields.

mgold:
This is a redo of #1324: detecting duplicate fields in record types, whether in annotations or aliases. The detection happens during validation, so there's a very nice error and it's possible to get more than one at a time.

Example message:

```
-- DUPLICATE FIELD ---------------------------------------------------- test.elm

This record type has more than one field named `foo`.

9│ g : List { foo : String, foo : Int } -> Int
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
There can only be one. Do some renaming to make sure the names are distinct!
```

Notice that it detected the duplicate even when nested in a list in a function, and also provides the exact location. This also works for aliases:

```
23│ type alias Foo = List { foo : String, foo : Int } -> Int
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The implementation is adding another value constructor to `Syntax.Error` an detecting it in `Validate`. As we've seen, we need to recursively traverse raw types to ensure there are no invalid records hiding.

Adds two tests. Renames the value constructor for record _values_ with duplicate fields for matching.

### DIFF
--- a/src/AST/Expression/Valid.hs
+++ b/src/AST/Expression/Valid.hs
@@ -28,3 +28,7 @@ getPattern :: Def -> Pattern.Raw
 getPattern (Def pattern _ _) =
   pattern
 
+
+getAnnotation :: Def -> Maybe Type.Raw
+getAnnotation (Def _ _ annotation) =
+  annotation
--- a/src/Reporting/Error/Syntax.hs
+++ b/src/Reporting/Error/Syntax.hs
@@ -28,7 +28,8 @@ data Error
 
     | InfixDuplicate String
     | TypeWithoutDefinition String
-    | DuplicateFieldName String
+    | DuplicateFieldNameInRecord String
+    | DuplicateFieldNameInRecordType String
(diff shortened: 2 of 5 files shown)


mgold:
Unfortunately the changes that just got merged in from dev mean this need to get reworked a bit. I'll get to it in a day or so.

evancz:
In my development branch, I am getting the following error messages for those cases:

```
-- NAME CLASH --------------------------------------------------------- temp.elm

This record has multiple `foo` fields.

4| g : List { foo : String, foo : Int } -> Int
              ^^^           ^^^
How can I know which one you want? Rename one of them!
```

```
-- NAME CLASH --------------------------------------------------------- temp.elm

This record has multiple `foo` fields.

9| type alias Foo = List { foo : String, foo : Int } -> Int
                           ^^^           ^^^
How can I know which one you want? Rename one of them!
```

The "duplicate name" errors should be much more precise in general though!

PR #1361, opened 2016-05-03
Update Tuple0 reference in Generate.JavaScript.Foreign

lukewestby:
I found that generating an incoming port with `()` as the incoming type made a reference to a variable `_Tuple0` which doesn't exist because of the new native code generation style in 0.17. I found this line here in the compiler and updated it with @eeue56's help to reference `Tuple0` in Native Utils.

(I'm having difficulty getting Haskell set up at the moment, and will reply back if I find that this change is no good)


lukewestby:
Closing: I haven't signed the contributor agreement yet

Issue #1363, opened 2016-05-05
Error message contains strange characters after line numbers


emtenet:
This happens on Windows 10 (Pro 64-bit and Home 32-bit) using 0.17-rc2 in a Command Prompt (cmd.exe) or PowerShell

This did not happen with 0.16 on the same systems.

Sample output from 0.17-rc2 on Windows 10 Pro 64-bit
```
d:\Source\Elm\Wired>"c:\Program Files (x86)\Elm Platform\master\bin\elm-make.exe" src\Error.elm
-- SYNTAX PROBLEM ------------------------------------------------ src\Error.elm


I ran into something unexpected when parsing your code!

3Ôöé type Error = invalid
                ^
I am looking for one of the following things:

    a constructor for a union type
    whitespace

Detected errors in 1 module.
```

The strange characters `Ôöé` are always the same for different errors or when an error displays multiple source lines.

Sample output from 0.16 on the same system
```
d:\Source\Elm\Wired>"c:\Program Files (x86)\Elm Platform\0.16\bin\elm-make.exe" src\Error.elm
-- SYNTAX PROBLEM --------- elm-stuff\packages\elm-lang\core\4.0.0\src\Array.elm


I ran into something unexpected when parsing your code!

1│ module Array exposing
                ^
I am looking for one of the following things:

    a listing of values and types to expose, like (..)
    reserved word `where`
    whitespace
```

jvoigtlaender:
That happens because the symbol in question is not available in the utf8 code page on Windows. It probably shouldn't be used at all. The reason why it is different between Elm 0.16 and Elm 0.17 is that on Elm 0.16 the compiler (or `elm-make`) didn't automatically switch to the utf8 code page. So in Elm 0.16 the situation was:

* If you don't switch the code page yourself, you do get to see the nice symbol there, but you can have massive problems like https://github.com/elm-lang/elm-make/issues/33 and https://github.com/elm-lang/elm-make/issues/40.
* If you do switch the code page yourself, by first calling `chcp 65001` in the shell before doing any Elm stuff, those problems are solved, but you also get to see the strange symbol in the error message that you are now seeing with Elm 0.17.

In contrast, the situation with Elm 0.17 is that the equivalent of `chcp 65001` is done by Elm itself, so you always are in the second of the two situations above.

jvoigtlaender:
[...]

PR #1364, opened 2016-05-06
Fix symbol following line number in error messages

jvoigtlaender:
Reason: At least on some Windows systems, the symbol previously used there does not show up properly when utf8 code page is active.

See https://github.com/elm-lang/elm-compiler/issues/1363.

### DIFF
--- a/src/Reporting/Render/Code.hs
+++ b/src/Reporting/Render/Code.hs
@@ -78,7 +78,7 @@ addLineNumber maybeSubRegion width n line =
       if n < 0 then " " else show n
 
     lineNumber =
-      replicate (width - length number) ' ' ++ number ++ "│"
+      replicate (width - length number) ' ' ++ number ++ "|"
 
     spacer (R.Region start end) =
       if R.line start <= n && n <= R.line end then


evancz:
I went in fixed this in the places it is needed. Did this happen on your computer? Or only on that other person's?

jvoigtlaender:
It happened on mine as well. Also what I wrote about 0.16.

evancz:
In the future, please mention it directly in issues like https://github.com/elm-lang/elm-make/pull/83

I don't know how you thought that was not relevant.

jvoigtlaender:
> I don't know how you thought that was not relevant.

Huh, what are you talking about? I didn't know that this happens (in 0.16 or later) until issue https://github.com/elm-lang/elm-compiler/issues/1363 was opened a few days ago. Then I experimented locally to diagnose that issue, and discovered that it is already the case in 0.16. So I provided the fix here, as well as pointing out over there what the cause was. Why would I go back to a closed issue (https://github.com/elm-lang/elm-make/pull/83) and add comments there about this there? Aren't we in the business of fixing problems in the here and now?

jvoigtlaender:
Also notice that https://github.com/elm-lang/elm-make/pull/83 was *not* a culprit here. The same "wrong symbol" occurred before https://github.com/elm-lang/elm-make/pull/83 (as I know now) when people did `chcp 65001` on Windows, as they had to in some situations to get the binaries work without throwing exceptions at all.

evancz:
It's fine, it's done. Next time you tell me ["I've tested this extensively"](https://github.com/elm-lang/elm-make/pull/83#issuecomment-196527379) just be more specific about what you did and did not test.

PR #1401, opened 2016-05-29
Safe var for outgoing ports

princejwesley:
Fixes https://github.com/elm-lang/core/issues/616 & #1414


YetAnotherMinion:
In #1414 referenced in this pull request @evancz says
> No longer a problem in 0.18

Should this pull request be closed?

evancz:
I guess so? If folks can verify that it is *not* fixed with 0.18, please open a new issue describing the problem.

PR #1407, opened 2016-06-06
Fix deps

evancz:


mgold:
TODO: tell @process-bot to shut it on PRs that Evan opens.

PR #1431, opened 2016-07-03
aeson-pretty: use Spaces with confIndent

ilovezfs:
Based on https://github.com/informatikr/aeson-pretty/blob/375afe8e/Data/Aeson/Encode/Pretty.hs#L124

Prevents the following error building Elm 0.17.1 with GHC 8.0.1:
```
    • No instance for (Num Json.Indent) arising from the literal ‘2’
    • In the ‘confIndent’ field of a record
      In the expression:
        Json.Config
          {Json.confIndent = 2, Json.confCompare = Json.keyOrder keys}
      In an equation for ‘config’:
          config
            = Json.Config
                {Json.confIndent = 2, Json.confCompare = Json.keyOrder keys}
            where
                keys = ["tag", ....]
```

### DIFF
--- a/src/Elm/Docs.hs
+++ b/src/Elm/Docs.hs
@@ -99,7 +99,7 @@ prettyJson value =
 config :: Json.Config
 config =
     Json.Config
-    { Json.confIndent = 2
+    { Json.confIndent = Json.Spaces 2
     , Json.confCompare = Json.keyOrder keys
     }
   where


jvoigtlaender:
That error would also be prevented by constraining `aeson-pretty` to a version `< 0.8` in the `.cabal` file.

ilovezfs:
And does this PR break it with that older version?

jvoigtlaender:
Yes.

ilovezfs:
Then that dep should be bumped.

jvoigtlaender:
That's for @evancz to decide.

ilovezfs:
I said it should be bumped, not that it will be ;)

evancz:
Since no one has confirmed that this code works with GHC 7.10.2 which I am using, I can't accept this PR for now. Not planning to do that kind of upgraded right now.

I have 0.7.2 of aeson-pretty, so if putting an upper bound on that package fixes it for GHC 8, I'd much rather go with that approach.

evancz:
Gonna go with the constraint instead. Thanks for bring the issue up and suggesting a fix!

gentoid:
Guys, look at this patch https://github.com/gentoo-haskell/gentoo-haskell/blob/master/dev-lang/elm-compiler/files/elm-compiler-0.15-aeson-pretty-0.8.patch from Gentoo Haskell overlay

PR #1439, opened 2016-07-15
Constrain version of aeson-pretty dependency

jvoigtlaender:
As requested.


evancz:
Ahh, I ended up doing the change real quick on the branch with the other changes. If both the things are green, I'll probably take the other one. Sorry for the communication mess up!

PR #1440, opened 2016-07-16
In issue template, direct to "meta" issues

jvoigtlaender:
The hope is that this will tremendously cut down on duplicates being opened.


mgold:
Huh, apparently GitHub markdown rendering makes links clickable without any square brackets. Maybe we should get rid of them from the other links.

Otherwise looks good to me.

PR #1442, opened 2016-07-19
Express unknown qualifier in terms of modules

rtfeldman:
This addresses https://github.com/elm-lang/error-message-catalog/issues/129 by framing the error in terms of how to fix it.

### Before

```
-- NAMING ERROR ------------------------------------------------------- form.elm

Cannot find variable `String.length`.

16|         if model.password == "" || String.length model.password < 8 then
                                       ^^^^^^^^^^^^^
The qualifier `String` is not in scope. 

Detected errors in 1 module.   
```

## After
```
-- NAMING ERROR ------------------------------------------------------- form.elm

Cannot find variable `String.length`.

16|         if model.password == "" || String.length model.password < 8 then
                                       ^^^^^^^^^^^^^
No module called `String` has been imported.

Detected errors in 1 module.   
```

### DIFF
--- a/src/Reporting/Error/Canonicalize.hs
+++ b/src/Reporting/Error/Canonicalize.hs
@@ -138,7 +138,7 @@ toReport localizer err =
           UnknownQualifier qualifier localName ->
               namingError
                 ("Cannot find " ++ var ++ ".")
-                ( "The qualifier `" ++ qualifier ++ "` is not in scope. "
+                ( "The `" ++ qualifier ++ "` module has not been imported. "
                   ++ Help.maybeYouWant (map (\modul -> modul ++ "." ++ localName) suggestions)
                 )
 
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
Well, how will this interact with something like:
```elm
import Module.Deep.Down as MyQualifier

...

function x = MyQualifier.name
```
The current error message "qualifier `MyQualifier` not in scope" makes more sense, then, than the new error message "`MyQualifier` module not imported". After all, such a thing as "the `MyQualifier` module" does not even exist.

jvoigtlaender:
Of course, the code in my previous comment wouldn't trigger the error at all, but assume I made some typo in one of the two occurrences of `MyQualifier`.

rtfeldman:
@jvoigtlaender Rephrased based on your feedback. New message would be:

```
No module called `MyQualifier` has been imported.
```

This is still a true statement, still easy for advanced users (who've used `as`) to understand the problem, and I think much more helpful for beginner cases like https://github.com/elm-lang/error-message-catalog/issues/129

jvoigtlaender:
Yes, I like this better. :smile:

evancz:
Really like the revised text. Thanks @rtfeldman and @jvoigtlaender!

PR #1444, opened 2016-07-19
Rephrase 'bad type annotation' error message

rtfeldman:
Rephrased "the type annotation for ______ does not match its definition," based on https://github.com/elm-lang/error-message-catalog/issues/110#issuecomment-233526116

This conveys the same information, but in a way that assumes (the more typical scenario where) the annotation they wrote was correct, but the implementation was broken.

## Before

```
-- TYPE MISMATCH ----------------------------------------------------- Main0.elm

The type annotation for `buggy` does not match its definition.

5│ buggy : List Int -> List Float
           ^^^^^^^^^^^^^^^^^^^^^^
The type annotation is saying:

    List Int -> List Float

But I am inferring that the definition has this type:

    List Float -> List Float

Hint: Elm does not automatically convert between Ints and Floats. Use `toFloat`
and `round` to do specific conversions.
<http://package.elm-lang.org/packages/elm-lang/core/latest/Basics#toFloat>
```

## After

```
-- TYPE MISMATCH ----------------------------------------------------- Main0.elm

The implementation of `buggy` does not match its type annotation.

5│ buggy : List Int -> List Float
           ^^^^^^^^^^^^^^^^^^^^^^
The type annotation is saying:

    List Int -> List Float

But I am inferring that the definition has this type:

    List Float -> List Float

Hint: Elm does not automatically convert between Ints and Floats. Use `toFloat`
and `round` to do specific conversions.
<http://package.elm-lang.org/packages/elm-lang/core/latest/Basics#toFloat>
```

### DIFF
--- a/src/Reporting/Error/Type.hs
+++ b/src/Reporting/Error/Type.hs
@@ -364,7 +364,7 @@ mismatchToReport localizer (MismatchInfo hint leftType rightType maybeReason) =
     BadTypeAnnotation name ->
         report
           Nothing
-          ("The type annotation for " ++ Help.functionName name ++ " does not match its definition.")
+          ("The implementation of " ++ Help.functionName name ++ " does not match its type annotation.")
           ( cmpHint
               "The type annotation is saying:"
               "But I am inferring that the definition has this type:"


evancz:
Merged in via 505bafcb6a73d7ba26ef5ab84e67b0ef7ff92a58

PR #1447, opened 2016-07-24
Compiler benchmarking

justinmanley:
Add a script for measuring how long it takes to compile Elm programs.

See [compiler-benchmark/README.md](https://github.com/justinmanley/elm-compiler/blob/compiler-benchmark/compiler-benchmark/README.md) for usage instructions.

I'd like to get feedback on what additional compilation phases might be useful to track, and if I'm tracking the ones I've chosen in the right way. See [`costCentreNames` in Benchmark.hs](https://github.com/justinmanley/elm-compiler/blob/compiler-benchmark/compiler-benchmark/Benchmark.hs#L88).

Here's an example of the output from running the script:

```
+----------------++------------------------------+
|                || %time in stable %time in dev |
+================++==============================+
| exhaustiveness ||               -         27.5 |
|   optimization ||               -          0.6 |
|        parsing ||               -         53.3 |
| type_inference ||               -         11.1 |
+----------------++------------------------------+
```
(There are no numbers reported for the stable version of the compiler because v0.17.1 does not have `{-# SCC #-}` annotations. This is WAI.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@ cabal-dev
 cabal.sandbox.config
 .DS_Store
 *~
+
+compiler-benchmark/benchmark-projects
+compiler-benchmark/benchmark-results
+compiler-benchmark/benchmark-src
--- a/compiler-benchmark/Benchmark.hs
+++ b/compiler-benchmark/Benchmark.hs
@@ -3,11 +3,17 @@ module Main where
 import Control.Monad (unless, forM_, void, when)
 import Data.Attoparsec.Text (parseOnly, maybeResult)
 import Data.IntMap ()
+import Data.List (zipWith, partition)
 import qualified Data.Text.IO as Text
-import Data.Text as Text (Text, pack)
+import Data.Text as Text (Text, pack, unpack)
 import Data.Foldable (find)
+import Data.Map (Map)
+import Data.Maybe (isJust)
+import Text.Tabular (Table(Table), Header(Header, Group), Properties(NoLine, SingleLine, DoubleLine))
+import qualified Text.Tabular.AsciiArt as Table
+import qualified Data.Map as Map
 import GHC.RTS.TimeAllocProfile (timeAllocProfile, profileHotCostCentres) 
-import GHC.RTS.TimeAllocProfile.Types (CostCentre, TimeAllocProfile, costCentreNodes, profileCostCentreTree, costCentreName)
+import GHC.RTS.TimeAllocProfile.Types (CostCentre, TimeAllocProfile, costCentreNodes, profileCostCentreTree, costCentreName, costCentreIndTime)
 import System.Directory (doesDirectoryExist, copyFile, setCurrentDirectory, getCurrentDirectory, createDirectoryIfMissing, getDirectoryContents, removeDirectoryRecursive)
 import System.Exit (ExitCode(ExitSuccess))
 import System.FilePath ((</>))
@@ -43,7 +49,7 @@ golden =
         , branch = "master"
         }
     ]   
-
+{-
 dev :: [Repo]
 dev =
     [ Repo
@@ -52,6 +58,7 @@ dev =
         , branch = "master"
         }
     ] 
+-}
 
 -- Names of annotated cost centres in elm-compiler and elm-make. Must be kept
 -- in sync with the SCC annotations in elm-compiler and elm-make (which must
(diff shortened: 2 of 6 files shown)


evancz:
Great, very excited to try this out and see what happens! I'm trying to focus on guide and a blog post at the moment, so I just wanted to explicitly let you know that this is on the work queue!

evancz:
As an aside, these numbers are already *extremely* interesting and useful! I can't think of any additional cost centers off the top of my head though.

I believe there's an issue open about improving the exhaustiveness check performance, and this indicates that that project is probably the lowest-cost highest-impact thing we can do right now.

evancz:
Alright, going to try to review today!

evancz:
Alright, I merged this in as the `compiler-benchmark` branch. I also added you to an @elm-lang team that has the ability to work on stuff like this in this repo. I have a couple comments though:

### Racing

One of the things I'm most keen to do is race 0.17.1 against master. So if `elm-todomvc` needs some minor changes from version to version, it also needs to be versioned. I think I just need to look around more to understand to race master against older versions.

I'm currently doing some work on the `dev` branch that should lead to some parser speed improvements for example, but I'm not sure how to run the race.

### Caching Profiling Binaries?

Maybe one way to facilitate races is to create profiling binaries like `elm-make-old` and `elm-make-new` so you can always compare to a concrete reference point. And as a developer, all you have to do is recreate the `elm-make-new` binary, which is necessary either way.

### Style

I want the Haskell code we write to look roughly [like this](https://gist.github.com/evancz/0a1f3717c92fe71702be), but the code is already not super consistent because much of it was written before this and hasn't been revisited in a while. Your code is really nice to read, but I just wanted to mention this overall consistency goal.

* * *

In any case, this seems like a solid foundation for starting on more serious optimization passes! For me the next step is to try to race `dev` against `0.17.1`. What are next steps for you?

evancz:
Alright, after thinking about it a bit more:

  - It seems like the cost center annotations should be added to the `elm-compiler` repo, and to `elm-make` possibly. That can be merged into `master` right now.
  - The code for creating a set of profiling binaries maybe should live in [this build script](https://github.com/elm-lang/elm-platform/blob/master/installers/BuildFromSource.hs).
  - The code for actually running the benchmark maybe should be in its own repo? Partly, I don't want `cabal install` to take longer when building binaries, and having a separate `.cabal` file is the only simple way I know to achieve that. Partly, this would mean we could have files that describe how to run 0.17.1 or master or whatever.
  - About the frailty of adding cost center comments by name, is there a way to have certain agreed upon prefixes and filter by that? Like, all tags start with `elm-*`. And maybe we have priority levels, so more filtering can be done. Like `elm-0-parsing` and `elm-1-binop-parsing` so you could ask for zero and lower.

I feel like these thoughts are sort of missing the point. What do you want to do next, and are there ways I can help make that happen?

evancz:
### More about cost center tags

What if there were "levels" of the cost center, so you can have annotations like this:

  - `elm/parse`
  - `elm/parse/binop`
  - `elm/parse/list`
  - `elm/inference`
  - `elm/inference/unify`
  - ...

So if you ask for one level, you just get `elm/parse` and `elm/inference`. If you ask for two levels, you get everything. Etc.

This way we can use the same benchmark runner no matter how drastically the compiler changes. Just start the comments with `elm/` and it'll be fine.

justinmanley:
Great - thanks for the comments!

My immediate next steps are to:

- Submit cost center annotations to elm-compiler (and elm-make if any are required there). I like the "elm/" prefix scheme for cost center tags much better than what I have now, so I'll switch the annotations and the benchmarking code to use the slash-delimited scheme you suggested.
- Enable racing 0.17.1 against dev by supporting stable and dev versions of the Elm projects used for benchmarking.

I think it's a sensible plan to move the benchmarking code to a separate repository. Then, the Benchmark.hs script could use elm-platform/BuildFromSource.hs to build different stable versions of the compiler. Moving the benchmarking code out of elm-compiler would make it easier to benchmark different branches of elm-compiler (currently, to race dev against 0.17.1, you need to rebase compiler-benchmark onto dev, then run the script). FWIW, the "cabal install" build time isn't so much of an issue, since cabal supports "Benchmark" build targets which are run with "cabal bench", and which (I believe) aren't built unless benchmarking is enabled with `cabal configure --enable-benchmarks`.

So the plan is:

- Move the compiler-benchmark directory from elm-compiler into a new repository.
- Refactor Benchmark.hs to use `elm-platform/BuildFromSource.hs`. This will require changes to BuildFromSource.hs to support an `--enable-profiling` command-line option.

evancz:
Makes sense to me!

PR #1449, opened 2016-07-25
Proposed revision to issue template text

jvoigtlaender:
* Call more explicit attention to it.
* Clarify that not all the items there are about "commonly reported", but some are also about issues that simply belong elsewhere.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,4 +1,6 @@
-Thank you for reporting an issue with Elm! Before submitting, please make sure it's not a commonly reported issue:
+PLEASE READ ALL THIS BEFORE OVERWRITING IT:
+
+Thank you for reporting an issue with Elm! Before submitting, please make sure it's not a commonly reported issue or an issue that belongs elsewhere:
 
 - If you're getting a runtime error, make sure you don't have any recursively defined values (see #873).
 - If you're having trouble with type annotations in a let-expression, see #1214.


evancz:
I personally think the problem is too many words. We can put *maybe* three bullets with 2 - 4 words each in there and expect folks to read them.

I think putting something bold will help as well, but I don't think that'll solve it. Fewer words will work better IMO

evancz:
If they had designed this feature with any care, they might have made it so people have to check a few things before writing. The current implementation is pretty silly. They just realized they could piggyback this "feature" on the existing site, basically without doing any technical or design work.

evancz:
AND you can't click any links in there!!!! Making people copy/paste things into a URL is 100% going to result in fewer people doing it. No one's fault but GitHub for designing it silly.

Anyway, I did a revised version with fewer words and some big words.

evancz:
Also, I hope this explains why I was more into process bot than this feature.

jvoigtlaender:
Yeah, as in the other issue, I also don't know what's going to fix the overall process.

Here just a few comments on what you wrote above:

> AND you can't click any links in there!!!!

Well, you can press the "Preview" button, and then have clickable text.

> Also, I hope this explains why I was more into process bot than this feature.

But is the bot addressing the problem of duplicate issues? Or the problem of issues being opened in the wrong repository? The hope would have to be that someone who opens an issue, then gets a bot message, which contains information that implies their issue is a duplicate or misplaced, that that someone would then close their issue in response.

Well, I don't know. Maybe the shortened text improves things. Thanks for investing into that.

evancz:
No one will press the preview button :)

We'll see how this works. It may be possible to have process bot try to do some things to avoid duplicates, but I don't want to make that text any longer really. It's already pushing it IMO.

At some point I converted to the mindset of "be aggressive about closing issues" with the understanding that people will open things again in the future if there is a good reason.

So maybe one way to handle it is, in a very nice tone, be like "I think this is a duplicate. Can you look through the open and closed issues to find anything relevant? If no, let's open a new issue that follows these guidelines" CLOSE

evancz:
I believe GitHub lets you create stock responses, so I think this can be a two click thing for folks who are looking through issues a lot. Maybe that's a plausible way to improve things?

jvoigtlaender:
> ... canned issue responses ...

Never tried, but might give it a try.

Issue #1457, opened 2016-08-08
Elm compiler hangs


z5h:
Running elm 0.17.1 on OSX.

The following (incorrect) code hangs the elm compiler.

```Elm
type SL comparable = SL (Set comparable) (List comparable)

updateSL : SL comparable -> comparable -> SL comparable
updateSL (SL set list) a =
  if (Set.member a set) then
    SL set list
  else
    SL (Set.insert a set) (list ++ a) -- should be (list ++ [a])
```

jvoigtlaender:
This is the same issue (just for different example code) as https://github.com/elm-lang/elm-compiler/issues/1456. My comment there applies here as well: It's the same issue as #1240, and as such is tracked in #1231.

@evancz, I wonder whether the issue template should contain an additional item of the form:
* Compiler hangs? https://github.com/elm-lang/elm-compiler/issues/1240

@z5h, if the issue template had contained such an item, would that have been enough for you to realize yours is a duplicate issue and you do not need to open it as a new one?

z5h:
@jvoigtlaender 1240 would have helped.

Issue #1240, opened 2015-12-23
compiler freezes compiling a recursive data type


folkertdev:
my compiler freezes when compiling the following snippet of code. This was reproduced by people in the elm irc channel. 

I am not sure that this code should compile (I may need List.map instead of concatMap), but in any case it shouldn't freeze the compiler. 

```elm
import Dict exposing (Dict)

type Trie comparable = Trie Bool (Dict comparable (Trie comparable))

toList : Trie comparable -> List (List comparable)
toList (Trie final dict) = 
    List.concatMap (\(key, trie) -> key :: toList trie) (Dict.toList dict)
```

system: Linux mint with elm 0.16, installed via npm.

In addition, the elm playground at elm-lang.org/try doesn't compile this snippet either, and displays a "server refused connection". Maybe some feedback can be given? (something like "The server could not compile your code")

A possible cause is the recursive data structure. Using a fold over this data structure to try get a List of elements also freezes the compiler. 

UPDATE

when the type signature is removed, this code does compile, but its type signature was incorrect all along. When changing `List.concatMap` to `List.map`, the compiler freezes. Other ways of arriving at this type signature also result in a compiler freeze.

```elm
toList : Trie comparable -> List (List comparable)
toList (Trie final dict) = 
    List.map (\(key, trie) -> key :: toList trie) (Dict.toList dict)
```

evancz:
It seems there are still cases where the compiler does not correctly catch infinite types. I have been tracking that #1231, so I'm going to close this issue and add a reference to it over there. This way I know to circle back to this one.

PR #1458, opened 2016-08-08
Add #1240 to issue template

jvoigtlaender:
As per https://github.com/elm-lang/elm-compiler/issues/1457#issuecomment-238225280.


evancz:
Thanks!

PR #1462, opened 2016-08-15
Feature improve emacs next error

aklaing:
Request to display error line number and column number to enable emacs next-error function.

### DIFF
--- a/src/Elm/Compiler.hs
+++ b/src/Elm/Compiler.hs
@@ -154,14 +154,14 @@ newtype Error =
     Error (A.Located Error.Error)
 
 
-errorToString :: Localizer -> String -> String -> Error -> String
-errorToString (Localizer localizer) location source (Error (A.A region err)) =
-    Report.toString location region (Error.toReport localizer err) source
+errorToString :: Bool -> Localizer -> String -> String -> Error -> String
+errorToString isEmacsStyle (Localizer localizer) location source (Error (A.A region err)) =
+    Report.toString location isEmacsStyle region (Error.toReport localizer err) source
 
 
-printError :: Handle -> Localizer -> String -> String -> Error -> IO ()
-printError handle (Localizer localizer) location source (Error (A.A region err)) =
-    Report.toHandle handle location region (Error.toReport localizer err) source
+printError :: Handle -> Bool -> Localizer -> String -> String -> Error -> IO ()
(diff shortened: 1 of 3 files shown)


evancz:
How come the emacs mode is not using `--report=json` which has this information available in an easier format? That seems like a better route than adding custom (and ugly) support to the compiler directly.

aklaing:
Hi Evan,

To be honest, I didn't think of it myself.  I don't know why jcollard and
bogdanp (two prominent elm-mode authors) didn't implement it that way.

Emacs assumes a model in which the compiler output will be placed in a
buffer just like a text file, and then it uses regular expressions to
search in that buffer for the error location information.  It is easy
enough to invoke the elm compiler with --report=json, but then you get a
purely machine-readable json dump (you know what that looks like).  While
emacs can correctly find the error information from there, all the benefits
of the elm-compiler's beautiful human-readable output are completely lost
for the human user.  Under normal usage with other compilers, that emacs
compilation-mode buffer serves two purposes -- provide the compiler
commentary to the user in a human readable fashion, and simultaneously
provide the "hotlinks" so that emacs can navigate to those regions in an
alternate window at the same time, side by side.  ie. Here's whats wrong
(from the compiler) and here's where you fix it, ....  Rendering the
"here's what's wrong" part in json would ruin everything.

No doubt emacs can be programmed to do two compile runs to get both forms
of information, or maybe it can be made to figure out how to produce elm's
normal nice output from the json.  However from a design point of view I
think those two alternatives are a lot less desirable and I'm guessing that
the impact on the emacs code base will be worse than the impact of my
suggested changes on the elm-compiler, to be completely honest.  Lastly, I
could not personally implement those two solutions.  I could implement a
simple solution that just parses the json but does not try to prettify it
for human readable purposes, but I would not even use that myself.  Not
when I can have proper elm-compiler messages :-).

I was hoping that by making it a compiler option, people who don't need it
will not see or be bothered by it.  Anyhow, those are the issues I can
think of.  I think this feature could benefit users of other editors than
emacs too, so the word "emacs" could be edited out of the elm-compiler code
base and a more editor-neutral word chosen.  My reference to emacs is
because it is the only editor I know that well.

Thanks!!

Ambrose

evancz:
Looks like we found an alternate route to get this functionality!

PR #1463, opened 2016-08-17
Fix wrong link in port type error

jvoigtlaender:
On the mailing list, [someone used a wrong port type and got this error](https://groups.google.com/forum/#!topic/elm-discuss/3ieqkWBPvaE):
```
8| port graphName : String
   ^^^^^^^^^^^^^^^^^^^^^^^
You are saying it should be:

    String

But you need to use the particular format described here:
<http://guide.elm-lang.org/effect_managers/>
```
However, the link http://guide.elm-lang.org/effect_managers/ doesn't describe any particular port type format. The link http://guide.elm-lang.org/interop/javascript.html does.


evancz:
Thanks!

Issue #1475, opened 2016-09-04
project names including `.` produce invalid generated code identifiers


michaelbjames:
Encountered while writing a new version of my website. I have the code in github under michaelbjames.com.git, clone to `michaelbjames.com/`

main.elm:

```
import Html exposing (text)

main =
  text "Hello, World!"
```

Will generate in the following invalid JS:
`var _michaelbjames$michaelbjames.com$Main$main = {
    main: _elm_lang$html$Html$text('Hello, World!')
};`

That pesky dot is what does it in. The elm code should work (it's taken from the most basic examples).

The fix is to better sanitize variable names before they're generated, so they don't conflict with the JS identifier spec.

Version information:
elm-compiler: master@4f5db922692099010a1cc5c987ce3cc536cf0140

Related task: https://github.com/elm-lang/elm-compiler/issues/1377

evancz:
In the latest version of the compiler, I reject project names with dots, so it never gets to this point.

Thanks for the report!

PR #1476, opened 2016-09-04
Sanitize dots and stars in github module names for JS

michaelbjames:
Resolves: https://github.com/elm-lang/elm-compiler/issues/1475

Previous bad thing:
```
var _michaelbjames$michaelbjames.com$Main$main = { main: _elm_lang$html$Html$text('Hello, World!') };
```

New good thing:
```
var _michaelbjames$michaelbjames_com$Main$main = { main: _elm_lang$html$Html$text('Hello, World!')
};
```

I'm sure this is not the only place where there's an inconsistency between the existing sanitizations and what ES5 is okay with. In fact, I doubt that just `-.*` is enough. 
* Do you think these swaps should map everything that is not `[a-zA-Z0-9$#_]` to `_`?
* Or should it use ecmascript's [`isValidIdPart :: Char -> Bool`](https://hackage.haskell.org/package/language-ecmascript-0.17/candidate/docs/src/Language-ECMAScript3-Syntax.html#isValidIdPart)
* Can you think of some better grand-plan for ensuring identifiers are valid?

### DIFF
--- a/src/Generate/JavaScript/Variable.hs
+++ b/src/Generate/JavaScript/Variable.hs
@@ -129,10 +129,10 @@ moduleToString :: ModuleName.Canonical -> String
 moduleToString (ModuleName.Canonical (Pkg.Name user project) moduleName) =
   let
     safeUser =
-      map (swap '-' '_') user
+      map (swapAny "-.*" '_') user
 
     safeProject =
-      map (swap '-' '_') project
+      map (swapAny "-.*" '_') project
 
     safeModuleName =
       List.intercalate "_" moduleName
@@ -144,6 +144,9 @@ swap :: Char -> Char -> Char -> Char
 swap from to c =
   if c == from then to else c
 
+swapAny :: [Char] -> Char -> Char -> Char
+swapAny badChars goodChar c =
+  if (c `elem` badChars) then goodChar else c
 
 
 -- SAFE NAMES


michaelbjames:
Poop, this didn't solve the bug for all cases I could come up with. I'll return with a better PR!

evancz:
You can have `*` in gihub repo names? My main worry is that we solve it but still allow certain problems. Maybe it is fairly safe to assume that no one will try to publish `foo-bar` as well as `foo.bar`? I kind of like the idea of doing a nice thing locally, but disallowing dots in the package repo. What do you think of that?

michaelbjames:
Looking at github's repo and user name validation: you can't have `*` in either.

I think disallowing dots will prevent this sort of issue affecting lots of users (if you only prevent them from publishing a package with a `.`) . But there should also be a safeguard in the compiler IMO. I'll use [isValidIdPart](https://hackage.haskell.org/package/language-ecmascript-0.17/candidate/docs/src/Language-ECMAScript3-Syntax.html#isValidIdPart) to swap out invalid characters for `_`.

So yeah, disallow dots, and I'll add this safeguard.

evancz:
I'd say just add the guard in [here](https://github.com/elm-lang/elm-compiler/blob/master/src/Elm/Package.hs#L105-L120). I'd do a special one for dots, because I assume they'd be common, but that `isValidPart` function seems like a nice way to solve this for good.

I kind of would like for things to mess up badly if it makes it past this check. That'd mean there's some way around it, which should be resolved, and I'd want to find that as early as possible.

michaelbjames:
Okay, I thought about it a little more:

Not permitting packages with dots **at all** makes me sad. Consider this semi-hypothetical case: I've made a repo 'michaelbjames.com' and put a few hundred commits in to it (because I was on redux). I now see the err of my ways and want to bring in Elm. However, now I'll have to rename my repo to use this sweet, new language. There could be varying levels of difficulty with that if you're not using git or your VCS is tightly controlled (at a company for instance). There are some pretty specific cases in that example but underlying it all:
It seems like an unnecessary gotcha to introduce when we could handle it in the compilation.

That said, I totally understand blocking dots in *published* packages. That could get funky to pass around. But local development seems like it should be less…regulated.

evancz:
When you are working locally, you can just give dummy names for the repo. The default is `https://github.com/user/project.git` and you never actually have to change it.

In the future, I think we will make it so you do not have to specify a repo unless you want to publish it, so it's even less of an issue.

So I think my proposed version does not clash with what you want. Am I correct?

Ailrun:
@michaelbjames Could you explain what cases aren't solved by this code?

imdaveho:
I have a similar case:
```
module Mainapp.Navigation.Types exposing (..)

import Mainapp.Navigation.Categories as Categories


-- TYPES
type AppState
    = ByCategory

type alias Store =
    { appstate : AppState
    , category : CategoryMenu
    }

type alias CategoryMenu =
    { category : Int
    , filter : List Int
    , subfilter : List Int
    , categories : Categories.Store
    }

type FetchMsg
    = Categories.Msg

type Msg
    = SelectApp AppState
    | SelectCategory Int
    | FetchCategories Categories.Msg

===================================
elm.js 
...
var _user$project$Mainapp_Navigation_Types$ByCategory = {ctor: 'ByCategory'};
var _user$project$Mainapp_Navigation_Types$Categories.Msg = {ctor: 'Categories.Msg'};
var _user$project$Mainapp_Navigation_Types$FetchCategories = function (a) {
	return {ctor: 'FetchCategories', _0: a};
};
...
```

notice the `Categories.Msg`

jvoigtlaender:
@imdaveho, yours is a different case, since it is not about Github names. But nevertheless, your case is already reported elsewhere in this repository.

jvoigtlaender:
More specifically, @imdaveho's case is exactly like this one: https://github.com/elm-lang/elm-compiler/issues/1200

PR #1488, opened 2016-09-16
move to current aeson-pretty (0.8.*)

thielema:


evancz:
Is there any rush on this? I am not targeting GHC 8 right now, and I expect to make a bunch of transitions like this when it's time for that. I'd rather do it all at once though.

amigalemming:
The problem is just that the released elm-0.17.1 fails to compile because version bounds on aeson-pretty are missing and thus Cabal tries to build Elm with latest aeson-pretty-0.8.*, which fails. I saw that version bounds fo aeson-pretty<0.8 were added, thus this might be enough for now but should be released.

PR #1481, opened 2016-09-08
reject dots in repo names

michaelbjames:
Resolves: #1475 

Should resolve discussion in prior PR: https://github.com/elm-lang/elm-compiler/pull/1476

PR #1501, opened 2016-10-11
Swap unsafes in variable name

Ailrun:
Solve https://github.com/elm-lang/elm-compiler/issues/1475 in different way

Ref: https://github.com/elm-lang/elm-compiler/pull/1481

In Github, including elm-lang repos, there are many repos with name having `.` character.
However, like `-` character, `.` makes invalid variable name in javascript.

At least I think, swapping these is more fundamental solution for dot problem.


Ailrun:
I didn't read https://github.com/elm-lang/elm-compiler/pull/1476. Sorry.
Discussion in the pull request is fair enough.

PR #1508, opened 2016-10-18
Link directly to #ports section for ports hint

rtfeldman:
Adds `#ports` to the end of the URL in the hint about ports.

https://guide.elm-lang.org/interop/javascript.html#ports takes you straight to the part of the page that explains the format.

**Error message before:**

```
But you need to use the particular format described here:

https://guide.elm-lang.org/interop/javascript.html
```

**Error message after:**

```
But you need to use the particular format described here:

https://guide.elm-lang.org/interop/javascript.html#ports
```

### DIFF
--- a/src/Reporting/Error/Canonicalize.hs
+++ b/src/Reporting/Error/Canonicalize.hs
@@ -266,7 +266,7 @@ toReport localizer err =
             , indent 4 (RenderType.toDoc localizer tipe)
             , Help.reflowParagraph $
                 "But you need to use the particular format described here:\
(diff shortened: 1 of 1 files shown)


evancz:
Makes sense, thanks!

jvoigtlaender:
Makes me wonder why I had to do [this removing of `#ports` parts](https://github.com/elm-lang/elm-lang.org/commit/4fe1ad37a8ac2d3e6c12b900d1aa617ec9ccf336), at [Evan's request](https://github.com/elm-lang/elm-lang.org/pull/501#issuecomment-218547320). :smile:

PR #1517, opened 2016-11-02
Remove unnecessary > in bad_recursion.md

mgold:
This is causing the line show show up indented and gray. Pretty sure you didn't want that!


mgold:
Looks better now, thanks!

PR #1553, opened 2017-01-08
Fix "need N more arguments" error message

spikefoo:
When compiling the following file:

```
foo = not || False
```

The given error message contains the line:

```
Hint: It looks like a function needs 1 more1 argument.
```

This mistake was introduced in commit 2feb3573, and only affects git master.


evancz:
I had to rebase to get it to merge into `dev`, so the commit is https://github.com/elm-lang/elm-compiler/commit/da3755abe598e646ca016903f3e41f5a49eab4cb

Thanks for reporting the mistake and providing a fix!

PR #1571, opened 2017-03-06
Implement task ports

eirslett:
`port myTask : Param -> Task Error Result`

will make a port where you can send a query to JS
and it will respond with a value or an error.

A proof-of-concept related to this discussion: https://groups.google.com/forum/#!topic/elm-discuss/TjWoacZobWw

Also requires changes to elm-core: https://github.com/eirslett/elm-core/tree/feature/task-ports

The implementation introduces a new type, `TaskPort` with an effect `Twoway`.
A port is determined to be a task port if it's on the form `a -> Task err res`. Similar to the way input/output ports are determined.


evancz:
I have talked with folks about why I do not think this is a good direction in the past. I think there is also discussion of it on the Elm Town podcast about "The History of Ports"

Separately, it is very important to discuss important language changes on the elm-dev mailing list before spending a bunch of time writing code. It doesn't matter if the code is great if the design is not going to be in the language.

PR #1596, opened 2017-05-23
Bump year in license

PeterUstinox:
same procedure as every year…


evancz:
I've seen FaceBook use `20XX-present` so I am going to go that route.

Thanks for pointing this out though!

PR #1616, opened 2017-07-18
Change non-Haddock comments which collide with Haddock comment syntax

tac-tics:
I tried building Haddock documentation for elm-compiler (later to find out there is very little currently). 

I noticed the Haddock build failing and the culprit was this file which made reference to the operators `|`, `*`, and `^` in the comments. I padded them so that Haddock builds properly, opening up that possibility to use it more in the future.


evancz:
The compiler is not set up to work with Haddock, so we can fix these, but it'll just break again later.

Creating good documentation in the compiler could be valuable for a small number of folks, but I do not think it makes sense to approach this in a piecemeal way with no overall goal or strategy.

Issue #1375, opened 2016-05-12
Tracking parser suggestions


evancz:
**Comments on this issue will be deleted. This is meant to organize problems in a coherent way.**

Known issues:

  - [ ] Remove `(,,)` syntax in types - #1236
  - [ ] Disallow whitespace in unit - #1210
  - [ ] Disallow whitespace in `( + )` - #1213

Issue #1068, opened 2015-09-24
Record ideas


evancz:
**This is a meta issue only for collecting related things in one place. If you think something is relevant, do not comment here, open a separate issue.**

People often ask about certain things, so I am collecting all the common requests in one place. I don't expect to be making any major changes here right now. This is more for book keeping.

  - ~~#985 - remove extension and deletion~~
  - #842 - allow more JS-ish style for closing }
  - #635 - allow qualified records in update syntax
  - #537 - allow arbitrary expressions in update syntax
  - #1254 - allow arbitrary types in type extension syntax

Apanatshka:
Maybe add #984 to the list, since it's waiting on 0.16 and #985?

jvoigtlaender:
Maybe add https://github.com/elm-lang/elm-compiler/issues/1254 to the list?

evancz:
Moved all this into #1375

jvoigtlaender:
How about @Apanatshka's proposal to add #984 as well? Seems lost in #1375.

evancz:
I considered that and decided against it. I know that proposal. It's a language feature proposal, not a simple syntax thing.

PR #1622, opened 2017-08-02
Expand record update syntax to allow general expressions before the p…

allanderek:
…ipe. Disallow empty updates such as '{ myRecord | }'.

I have added a test file tests/test-files/good/Records, but not a corresponding one in tests/test-files/good-expected-js/, the reason was based on the fact that on the dev branch the compiler test suite was commented out. I uncommented it locally for testing, but only fixed up the types enough to get it to compile not to produce expected Javascript.

### DIFF
--- a/src/Parse/Expression.hs
+++ b/src/Parse/Expression.hs
@@ -189,33 +189,34 @@ tupleHelp start entries =
 record :: R.Position -> Parser Src.RawExpr
 record start =
   do  leftCurly
+      spaces
       inContext start E.ExprRecord $
-        do  spaces
-            oneOf
-              [ do  rightCurly
-                    end <- getPosition
-                    return (A.at start end (Src.Record []))
-              , do  starter <- addLocation lowVar
-                    spaces
-                    oneOf
-                      [ do  pipe
-                            spaces
-                            firstField <- chompField
-                            fields <- chompFields [firstField]
-                            end <- getPosition
-                            return (A.at start end (Src.Update (A.map Src.var starter) fields))
-                      , do  equals
-                            spaces
-                            (value, _, nextPos) <- expression
-                            checkSpace nextPos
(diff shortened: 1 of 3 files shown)


Skinney:
Does this mean that one could do:

    { model | nested = { model.nested | value = 5} }

?

allanderek:
Yes you can do that.

jvoigtlaender:
For reference, this is related to https://github.com/elm-lang/elm-compiler/issues/1068, https://github.com/elm-lang/elm-compiler/issues/1375.

evancz:
The important part of this feature is deciding if it is a good *design*, not writing code.

I did not add this feature because I did not think it was a good move and would lead to less readable code. Having a PR does not change any of the fundamental thinking on this, so I will close.

evancz:
This change has been proposed and discussed many times before. I will restate some of the design problems here.

When you say `{ person.position | x = 4 }` is the return type `Person` or `Point`? I am not asking this question. I am observing that every user of Elm will have to ask this question every time they see code like this.

Now if we have one that gives `Point`, the next feature request is for one that gives `Person` obviously. That will be the next PR after this. So do we want `{ person | position.x = 4 }` as well?

Okay, say we do not add that though. We just have the `{ <EXPR> | x = 4 }` syntax. I 100% guarantee you that one of your coworkers will put a multi-line expression in there. It will be a `let` or another record or a `case`. And they will tell you that it is actually easier this way. It reads more naturally. Someone out there will do a 30 line expression in there because that is their preference.

So in my opinion, the pursuit of a minor convenience here unlocks a huge amount of corner cases that **will** be common in practice. I think overall code quality will degrade. Without this change, you are encouraged to write helper functions. This is better for your code, better for teams, etc. So the "fix" here is not worth the unintended consequences in my opinion.

Sorry this was not clear from my comments elsewhere.

PR #1631, opened 2017-08-16
accept binary 0.8.*

imuli:
Required for building on current NixOS stable.

`cabal test` didn't bring up anything.

I don't think this patch is copyrightable (not a creative work) and consider it under the Public Domain, but I can get you a contributors agreement if you need it.


evancz:
The master branch accepts this now.

https://github.com/elm-lang/elm-compiler/blob/master/elm.cabal#L225

Got things upgraded to GHC 8.2 as well.

PR #1663, opened 2018-01-06
bump year in license

BenBals:


evancz:
The license in master got updated, I think when things were in `dev`.

Thank you for the fix though!

PR #1666, opened 2018-01-16
Misc typos and rewording

timbuckley:
Self-explanatory


evancz:
I believe these are all fixed in https://github.com/elm-lang/elm-compiler/commit/55a65a190750671f8ec3dc455855a6d6554f96ec and https://github.com/elm-lang/elm-compiler/commit/3075c39eb50346860e744b27226e2a60dba011ee

Thank you for the reports!

PR #1698, opened 2018-05-06
fix typo, genarate -> generate

brandly:
looks like @zwilias noticed it first: https://github.com/elm-lang/elm-compiler/commit/28037723ecd0e1ad21ca410a423343761ee1018e#commitcomment-27849371

thanks!


evancz:
I think I fixed it, but I'd rather just leave it for now if not. I'll notice it again at some point and fix it then :)

PR #1723, opened 2018-06-15
Update core library names from elm-lang/* to elm/*

Edward-Lombe:


evancz:
I believe this got changed at some other time.

evancz:
Oh, that was the application one! I changed this one as well now.

PR #1726, opened 2018-06-16
update application elm.json example

ryannhg:
Hey Evan!

While `elm init` is still a work in progress, this resource is really useful for getting started with an elm application!

I thought I should update the example based on the latest package names and versions.

Thanks,
Ryan


ryannhg:
Nevermind! Looks like the `elm.json` is in a new format now 😄

PR #1727, opened 2018-06-17
Fix corner style of boxes in reactor

jinjor:
I believe this fixes a style in reactor like this.

Before:
image

After:
image

Another plan would be adding `overflow: hidden;` (or scroll, auto ...) to `.box`.

```css
.box {
    border: 1px solid #c7c7c7;
    border-radius: 5px;
    margin-bottom: 40px;
+    overflow: hidden;
}
```


evancz:
I was wondering about that! Thank you!

Issue #1732, opened 2018-06-23
New lines are not escaped in the generated shader code


w0rm:
The following code:

```elm
module Main exposing (main)

import Browser
import Html exposing (Html)
import Math.Vector3 as Vec3 exposing (Vec3, vec3)
import WebGL exposing (Mesh, Shader)


main : Html a
main =
    WebGL.toHtml [] [ WebGL.entity vertexShader fragmentShader mesh {} ]

type alias Vertex = { position : Vec3 }

mesh : Mesh Vertex
mesh =
    WebGL.triangles
        [ ( Vertex (vec3 0 0 0)
          , Vertex (vec3 1 1 0)
          , Vertex (vec3 1 -1 0)
          )
        ]

vertexShader : Shader Vertex {} {}
vertexShader =
    [glsl|
        attribute vec3 position;
        void main () {
            gl_Position =  vec4(position, 1.0);
        }
    |]

fragmentShader : Shader {} {} {}
fragmentShader =
    [glsl|
        precision mediump float;
        void main () {
            gl_FragColor = vec4(0, 0, 0, 1.0);
        }
    |]
```

With the following `elm.json`:

```json
{
  "type": "application",
  "source-directories": ["."],
  "elm-version": "0.19.0",
  "dependencies": {
    "direct": {
      "elm-explorations/linear-algebra": "1.0.0",
      "elm-explorations/webgl": "1.0.0",
      "elm/browser": "1.0.0",
      "elm/core": "1.0.0",
      "elm/html": "1.0.0",
      "elm/json": "1.0.0"
    },
    "indirect": {
        "elm/time": "1.0.0",
        "elm/url": "1.0.0",
        "elm/virtual-dom": "1.0.0"
    }
  },
  "test-dependencies": { "direct": {}, "indirect": {} }
}
```

Produces the following JavaScript:

```javascript
var author$project$Main$vertexShader = {src: '

        attribute vec3 position;

        void main () {
            gl_Position =  vec4(position, 1.0);
        }

    '};
```
 
Possible fix: https://github.com/elm/compiler/pull/1733 outputs: 

```javascript
var author$project$Main$vertexShader = {src: '\n\n        attribute vec3 position;\n\n        void main () {\n            gl_Position =  vec4(position, 1.0);\n        }\n\n    '};
```

evancz:
Thanks for the report!

PR #1733, opened 2018-06-23
Escape new lines in shader strings

w0rm:
A possible fix of  #1732.


evancz:
The fix I committed escapes things a bit more, and does it earlier in the pipeline. So the escaping should only happen once, not on every code gen.

PR #1756, opened 2018-08-22
Improve wording for package build cache explanation

Razzeee:
Hey there,

I'm not sure if you wanted that sentence to sound like that or if my idea is any better. But I think we can improve it.

I used the github editor, and for some reason it shows the last lines as changed too. So if this goes anywhere I will fix that.


evancz:
I think it is more clear the existing way. When you start a fresh project (a brand new project) it will use the cached package. So a package is only built once. It's not per-project. It is per-user.

Razzeee:
Well was more thinking in line of the user that just jumped to elm 0.19 or a user that just installed elm. In that case the description is wrong, as the project is fresh and will still need to build the deps.

Anyway, I don't think changing this would be helpful as it might cause more questions and lead to a worse wording then keeping it as is.

PR #1775, opened 2018-08-30
Remind those making issues that code should be in issue directly, not just Ellie links

timbuckley:
Per Evan's comment here: https://github.com/elm/compiler/issues/1762#issuecomment-417419430


evancz:
The latest version of the template tries to do this in a different way:

https://github.com/elm/compiler/blob/master/.github/ISSUE_TEMPLATE.md

I got the impression that people would just `Ctrl-a` `Backspace` with all the words that were in there, so I accepted that the point of the tool is just to prefill some text, not to communicate any additional information that requires reading.

PR #1783, opened 2018-09-03
Export most modules as library

reactormonk:
For IDE efforts via LSP, so forks like

https://github.com/norpan/elm-compiler-library/commit/c1607fd1cd5a98f6fbc74c3407b57938bb97964d

Are not required anymore


evancz:
This will 2x compile times for compiler work.

It seems plausible to swap the `.cabal` file out with whatever you want. That seems less invasive for the relevant parties. I have very little information to go on from the messages here though.

reactormonk:
> This will 2x compile times for compiler work.

That can be fixed with a bit more work.

> It seems plausible to swap the .cabal file out with whatever you want.

Yeah, that's the current solution. Requires fork maintenance though.

> That seems less invasive for the relevant parties.

Aka the compiler developers? Would be easier for the IDE developers to have a library available and not have to maintain their own fork. Also gives you the ability to have a test suite as part of the compiler.

> I have very little information to go on from the messages here though.

From what I get, you're not interested in a PR in this direction, given you closed this one. I don't mind, @norpan is the guy working on the LSP stuff and he's fine with the fork.

PR #1784, opened 2018-09-03
Relax maximum language-glsl version to fix build with ghc-8.4

rlefevre:
This is needed to be able to build `elm` with `ghc-8.4/base-11.4` (tested with `ghc-8.4.3`):

See https://github.com/noteed/language-glsl/pull/17#issuecomment-417923838 (version 0.3.0 released yesterday with the fix)

Maybe some tests should be performed to be sure that `elm-explorations/webgl` still work as expected though.


evancz:
I recently got everything compiling with 8.4 locally, so this will be changed through that work.

PR #1814, opened 2018-09-30
Make reactor work with single page applications

dbdbdb:
This let you pass in an elm file to the reactor command so it can be used to serve a single page application. Files that exists on disk (and elm assets) are served first and other urls are passed to the SPA so it can do its own routing.


ryanolsonx:
I'd love to have this in elm reactor!

dbdbdb:
This is almost three years old now so I'll close it because of that.

PR #1829, opened 2018-10-18
fix mistaken repeat of word "matching" in error message

jerith666:


evancz:
Thank you for reporting this! The "merge" button on my side was red and I wasn't sure what the implications would be of pressing it, so I just made the changes locally to be cautious.

PR #1834, opened 2018-11-01
Add more examples useage for the output flag

razzeee:
This adds some more explanation on how to generate html. As it wasn't mentioning the html output before.

Hope this omission was not on purpose.


Razzeee:
Rebased

Razzeee:
Rebased

PR #1841, opened 2018-11-14
Allow elm to be installed via npm on node 10

nono:
See https://github.com/avh4/binwrap/issues/11

### DIFF
--- a/installers/npm/package.json
+++ b/installers/npm/package.json
@@ -39,6 +39,6 @@
     "elm": "bin/elm"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


evancz:
We decided not to rely on `binwrap` anymore due to some problems with spurious security warnings. The necessary functionality is now implemented within the `elm/compiler` project such that there are no external `npm` dependencies. This new approach should be available via `0.19.0-no-deps` right now, and will be available with some improvements once 0.19.1 is ready for release.

PR #1842, opened 2018-11-15
fix typo in error message: s/piping it a function/piping it to a func…

jerith666:
…tion/


evancz:
There were some weird notes in GitHub about pressing the merge button, I think because the ongoing work in the compiler has led to a bunch of files getting moved around. I just added the missing word locally to get the fix instead. Thank you for pointing this one out!

PR #1847, opened 2018-11-19
Replace "union type" with "custom type"

rtfeldman:
Also replace "constructor" with "variant" and replace link to `union_types.html` with link to `custom_types.html`


rl-king:
Also a fix for: https://github.com/elm/error-message-catalog/issues/271

PR #1850, opened 2018-11-19
Fix Map.!: given key is not an element in the map

tomstejskal:
I have an application using the package [mdgriffith/elm-ui](https://package.elm-lang.org/packages/mdgriffith/elm-ui/latest/).
One of the application `Msg`data constructors contains value of type `Element msg` from the elm-ui library.
Declaration of this type is:
```elm
type alias Element msg =
   Internal.Element msg
```
Neither the `Internal.Element msg` union type nor the `Internal` module is exposed from the library.
This pull request fixes function `extractUnion`, which couldn't find such module and type.

I'm not sure this is the right way how to solve this issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,10 @@
 elm-stuff
 dist
+dist-newstyle
 cabal-dev
 .cabal-sandbox/
 cabal.sandbox.config
 .DS_Store
 *~
+.ghc*
+cabal.project.local
--- a/builder/src/Deps/Website.hs
+++ b/builder/src/Deps/Website.hs
@@ -29,7 +29,7 @@ import qualified Network.HTTP.Client as Client
 import qualified Network.HTTP.Client.MultipartFormData as Multi
 import qualified Network.HTTP.Types as Http
 import qualified System.Directory as Dir
-import System.FilePath ((</>), splitFileName)
+import System.FilePath ((</>), takeDirectory, splitFileName, hasTrailingPathSeparator)
 
 import Elm.Package (Name, Version)
 import qualified Elm.Package as Pkg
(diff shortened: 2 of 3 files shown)


tlentz:
Merge this, it works and fixes the elm-deugger.

domenkozar:
Potential fix for https://github.com/elm/compiler/issues/1851

Someone could confirm if it fixes SSCCE in https://github.com/elm/compiler/issues/1753#issue-352971149

Natim:
It fixes #1851 can we release this in elm 0.19.1, it would greatly help me to be able to use the debugger.

domenkozar:
@natim have you confirmed the fix?

Natim:
image

Natim:
> @Natim have you confirmed the fix?

Yes I do :+1: 

```
$ npm run debug

> catalog@0.0.1 debug ~/elm/catalog/ui
> elm-live src/Main.elm --port=3000 --dir=public/ -- --output=public/app.js --debug

Success! Compiled 1 module.                                          

elm-live:
  The build has succeeded. Starting the server!


Reload web server:
  - Website URL: http://localhost:3000
  - Listening on port: 3000
  - Monitoring dir: public/

elm-live:
  Watching src/**/*.elm.
```

JDemler:
We merged this PR into [our fork](https://github.com/CurrySoftware/compiler) and extended the reactor so that is capable of compiling with the `--debug` flag.

jesse-c:
Confirmed to be working for me as well (macOS)

Maxim-Filimonov:
Confirmed working for @JDemler  fork on  OSX using our [test repo](https://github.com/Maxim-Filimonov/Elm-graphql-YT).
Compiles and runs as expected with the time travelling debugger 👍

domenkozar:
cc @evancz

ocharles:
Another confirmation that this fixes the issue for me - before we couldn't build our application with `--debug`, but with this patch applied we can.

domenkozar:
@evancz there are requests making `--debug` optional in parceljs, I'd like to avoid that and rather fix the c ompiler with a bugfix release. How realistic is that?

domenkozar:
doesn't seem this bug wil get fixed until 19.1 https://github.com/elm/compiler/commit/af2ea2b5baf6e47dcccca4385f6a773a981c6df2

francescortiz:
> doesn't seem this bug wil get fixed until 0.19.1 [af2ea2b](https://github.com/elm/compiler/commit/af2ea2b5baf6e47dcccca4385f6a773a981c6df2)

Any idea when it is happening? Just showing which package triggers the error would be a already great improvement.

domenkozar:
@evancz I've hit this bug again today.

It would be great to have feedback if it will be fixed in 0.19 or not, to engineer our codebases accordingly.

evancz:
This patch adds some logic to avoid the `Map.!` error, but it does so by hiding the underlying problem.

The root issue was that the types for all package dependencies should be available to the compiler when generating the `--debug` mode to make sure that types have not changed, so as part of the ongoing structural changes in the compiler, I introduced the `DependencyInterface` type which makes sure the relevant information from transitive dependencies is available. From there, I made some additions like the `Extract.Types` type to document the need for this information more explicitly in the structure of the code itself.

- https://github.com/elm/compiler/blob/master/compiler/src/Elm/Interface.hs#L156-L161
- https://github.com/elm/compiler/blob/master/compiler/src/Elm/Compiler/Type/Extract.hs#L88-L89
- https://github.com/elm/compiler/blob/master/compiler/src/Generate/Mode.hs#L24-L26
- https://github.com/elm/compiler/blob/master/builder/src/Generate.hs#L180-L188

I think these changes are a better solution to the root problem. They will be available in the 0.19.1 release, so I am going to close this PR but keep the issues about this bug open at least until the fix is confirmed in the alpha period for the next release.

m-mullins:
For 0.19.0 my issue was solved by moving `elm/virtual-dom` to a direct dependency.

PR #1876, opened 2019-01-03
Fix typo in error message: multilpe -> multiple

paulfioravanti:


evancz:
Thank you for the fix! I ended up making the changes locally since a lot of things have moved around in the project and something was causing a merge conflict such that it wasn't possible to just press the "merge" button on this.

PR #1884, opened 2019-01-18
Update npm installer to binwrap 0.2.0

rtfeldman:
This is already published on npm; just making a PR to update `elm/compiler` to reflect reality!


harrysarson:
What would need to happen for this PR to be merged?

evancz:
Ended up deciding to implement this without any dependencies at all. Trying to minimize the vectors by which users can get irrelevant security warnings about risky packages, and having zero dependencies seems like the best bet.

PR #1886, opened 2019-01-21
fix writeArchive to use Zip.fromEntry rather than Zip.writeEntry

### LINKED ISSUE #1882: elm install - UnsafePath error on install from zipball
Hi,

Elm version 0.19.0, macOS 10.13.6.

Command `elm install 1602/elm-feather` was failing with:

```bash
 -- HTTP PROBLEM ----------------------------------------------------------------

The following HTTP request failed:

    <https://github.com/1602/elm-feather/zipball/2.3.4/>

Here is the error message I was able to extract:

    UnsafePath "/Users/tzemanovic/.elm/0.19.0/package/1602/elm-feather/2.3.4/"
```

I found that the condition in zip-archive for this is:
```haskell
let isUnsafePath = ".." `isInfixOf` absPath ||
                     not (curDir `isPrefixOf` absPath)
```

From there I found a work-around to `elm init` and `elm install` this package in `~/.elm/0.19.0` directory, which indeed worked and then I was able to install it elsewhere thanks to cache.

jerith666:
avoids UnsafePath error with zip-archive >= 0.4

fixes #1882 

---

I've tested that this produces the same `~/.elm` and `elm-stuff/` directory structure as well as the same `.js` file with and without optimization, on two of my Elm projects.

I have not tested with zip-archive < 0.4.

### DIFF
--- a/builder/src/Deps/Website.hs
+++ b/builder/src/Deps/Website.hs
@@ -29,7 +29,7 @@ import qualified Network.HTTP.Client as Client
 import qualified Network.HTTP.Client.MultipartFormData as Multi
 import qualified Network.HTTP.Types as Http
 import qualified System.Directory as Dir
-import System.FilePath ((</>), splitFileName)
+import System.FilePath ((</>), takeDirectory, splitFileName, hasTrailingPathSeparator)
 
 import Elm.Package (Name, Version)
 import qualified Elm.Package as Pkg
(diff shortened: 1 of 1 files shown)


evancz:
As part of the ongoing structural changes in the compiler, I ended up making similar changes to the code for downloading zip files:

https://github.com/elm/compiler/blob/master/builder/src/File.hs#L164-L185

It also skips `Zip.writeEntry` but in a way that is specialized to the specific needs of this logic in the compiler.

PR #1889, opened 2019-01-29
Fix contributor agreement link

davetapley:
GitHub is expanding the file name to `https://github.com/elm/compiler/blob/master/.github/ContributorAgreement.pdf`, which is a 404, so use the full URL instead.


mskiptr:
This can be closed as the issue was first fixed in #2138 and then in efb6711757d40acee612f797c02cafca020cc073 the CLA process was thankfully abandoned altogether.

PR #1891, opened 2019-02-05
Update GitHub issue template links

asterite:
Hello!

This PR updates the links from the apparently old `elm-lang` organization to `elm`.

This is mostly superficial except for one link:

> Core libraries? https://github.com/elm-lang/core/issues

That page doesn't exist so you get redirected to the "pull requests" page of that repo. You have to go back to the root, go to the updated repo name and then go to issues. It's a bit confusing if you want to report an issue in the core library (which I wanted to do).

(Side note: I also found some other dead links in other `.md` files, for example the "stable" branch doesn't exist anymore... but I wanted to keep this PR focused)

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,11 +1,11 @@
 BEFORE DELETING THIS:
 
-- Runtime error? https://github.com/elm-lang/elm-compiler/issues/1591
-- Type annotations? https://github.com/elm-lang/elm-compiler/issues/1214
-- Core libraries? https://github.com/elm-lang/core/issues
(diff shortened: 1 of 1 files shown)


evancz:
Thank you for the fix! I generally use a batching technique for PRs, so I end up looking through them all in one go to see how they fit together with each other and with any independent changes due to ongoing work.

In this case, I got the sense that nobody would click any of these links since you have to copy/paste them to actually take a look at them. So at some point I switched the template to https://github.com/elm/compiler/blob/master/.github/ISSUE_TEMPLATE.md in an effort to at least get OS and Elm version information more often. I think this is the intended use of the templates, or at least the use that has more benefits in practice.

Issue #1736, opened 2018-06-25
Parsing regression in 0.19 `elm repl`


rtfeldman:
If you try to use `==` in a top-level `elm repl` expression, and don't wrap it in parens, you now get a parse error.

# SSCCE

```> 3 > 4
False : Bool
> 3 == 4
-- PARSE ERROR ------------------------------------------------------------- elm

Something went wrong while parsing your code.

3| 3 == 4
   ^
I was expecting:

  - a declaration, like `x = 5` or `type alias Model = { ... }`
  - the `import` keyword
  - a doc comment, like {-| this -}
> (3 == 4)
False : Bool
```

`3 == 4` works as expected in 0.18 `elm repl`.

Issue #1899, opened 2019-02-22
How to build Elm itself?


radrow:
I would like to contribute to the Elm compiler in the near future, so I wanted to build it on my machine locally. However, I cannot find any build instructions for it. I ran `cabal build`, but got a hell load of dependency errors that after fixing were generating even more of them (just like fighting a hydra). As advised by warning I used `cabal new-build`, but got build error in X509:

```
Data/X509/AlgorithmIdentifier.hs:14:1: error:
    Could not find module ‘Data.ASN1.Types’
    There are files missing in the ‘asn1-types-0.3.2’ package,
    try running 'ghc-pkg check'.
    Use -v to see a list of the files searched for.
   |
14 | import Data.ASN1.Types
``` 

With `cabal build` the problems are in packages `http-client`, `snap-core` and `snap-server`:

```
Downloading  http-client-0.6.1.1
Starting     io-streams-1.5.1.0
Building     io-streams-1.5.1.0
Downloaded   http-client-0.6.1.1
Starting     http-client-0.6.1.1
Building     http-client-0.6.1.1
Failed to install http-client-0.6.1.1
Build log ( /home/radek/.cabal/logs/ghc-8.6.3/http-client-0.6.1.1-2TabBoAYA1LGCkNuKzYIME.log ):
cabal: Entering directory '/tmp/cabal-tmp-8554/http-client-0.6.1.1'
Configuring http-client-0.6.1.1...
Preprocessing library for http-client-0.6.1.1..
Building library for http-client-0.6.1.1..
[ 1 of 19] Compiling Data.KeyedPool   ( Data/KeyedPool.hs, dist/build/Data/KeyedPool.o )
[ 2 of 19] Compiling Network.HTTP.Client.Types ( Network/HTTP/Client/Types.hs, dist/build/Network/HTTP/Client/Types.o )

Network/HTTP/Client/Types.hs:38:1: error:
    Could not find module ‘Network.HTTP.Types’
    There are files missing in the ‘http-types-0.12.2’ package,
    try running 'ghc-pkg check'.
    Use -v to see a list of the files searched for.
   |
38 | import Network.HTTP.Types
   | ^^^^^^^^^^^^^^^^^^^^^^^^^

Network/HTTP/Client/Types.hs:43:1: error:
    Could not find module ‘Blaze.ByteString.Builder’
    There are files missing in the ‘blaze-builder-0.4.1.0’ package,
    try running 'ghc-pkg check'.
[...]

PR #1903, opened 2019-02-28
Fix `==` operator parsing in REPL

radrow:
fixes: #1736 

Hi, I have reviewed mentioned issue which was just a simple parsing error caused by the fact that after parsing a symbol the parser didn't check whether the operator continues or not, which was leading to mistaking `=` with `==`. To solve the problem I check whether next character (if present) in the input right after the symbol belongs to the `binopCharSet`. If it is there - parsing fails like if we met non-matching char.

In my fix I have separated "operator" symbols from "section mark" symbols to allow use of `(+)` functions. It was necessary, because without this feature the "leftParen" would be rejected due to "binopChar" presence right after it.

As a side effect, I was forced to rename one local variable in `Parse/Primitives/Variable.hs` in order to compile, because it turned to be restricted id by GHC (https://wiki.haskell.org/Keywords#foreign). This identifier does not leak to the toplevel scope, so I hope it won't be a problem.

I was working on tag `0.19.0` as advised in the issue #1899 . I see there are some conflicts with the master – after the review I can resolve them and prepare to merge.

---------------

It is my first contribution to open source, I would be glad for any feedback and advises.

### DIFF
--- a/compiler/src/Parse/Primitives/Symbol.hs
+++ b/compiler/src/Parse/Primitives/Symbol.hs
@@ -17,7 +17,9 @@ import qualified Data.ByteString.Char8 as Char8
 import qualified Data.Char as Char
 import qualified Data.IntSet as IntSet
 import qualified Data.Vector as Vector
-import Foreign.ForeignPtr (ForeignPtr)
+import Foreign.Ptr (plusPtr)
+import Foreign.ForeignPtr (withForeignPtr, ForeignPtr)
+import Foreign.Storable
 import GHC.Word (Word8)
 
 import qualified Elm.Name as N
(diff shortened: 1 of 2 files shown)


evancz:
Thank you for the suggestion on how to resolve #1736!

There are a bunch of structural changes happening in the compiler at the moment, and one result of that was that the logic for the REPL was done a bit different, partly so that this issue would get resolved. So on my local build, I am now seeing:

```elm
---- Elm 0.19.1 ----------------------------------------------------------------
Say :help for help and :exit to exit! More at <https://elm-lang.org/0.19.1/repl>
--------------------------------------------------------------------------------
> 4 == 10
False : Bool
```

But the changes I made did not require any changes to the parser. It was all done externally. One thing that is risky when fixing a bug like this is introducing vectors for _other_ bugs. I think the approach that is used in the master branch has a better balance since changing the parser for Elm files (which is more important) will cause compile errors in the REPL. Whereas adding conditional features means that parsing Elm files now has one more thing to worry about when writing the code, and the compiler cannot help you if you get it wrong.

PR #1913, opened 2019-03-14
upgrade binwrap to drop deprecated natives dependency

michaelglass:
binwrap 0.2.0 doesn't have a transitive dependency on natives, which is a dicey package (See this note: https://github.com/addaleax/natives#caveat)

Specifically opening this PR because of test failures here https://github.com/NoRedInk/elm-css-files/pull/1

### DIFF
--- a/installers/npm/package.json
+++ b/installers/npm/package.json
@@ -39,6 +39,6 @@
     "elm": "bin/elm"
   },
   "dependencies": {
-    "binwrap": "0.1.4"
+    "binwrap": "0.2.0"
   }
 }


harrysarson:
See #1884, the package on npm has already been released with this version bump. The changes have not yet been merged into this repository though.

There is also a discourse post somewhere with more information.

jhrcek:
Here is the mentioned topic on discourse https://discourse.elm-lang.org/t/why-are-updates-to-the-official-npm-package-coming-out-of-an-obscure-branch-on-not-the-official-repo

PR #1940, opened 2019-06-18
Add fonts mime type mapping to elm reactor

albertdahlin:
Fonts have their proper mime types since Feb 2017.

https://www.iana.org/assignments/media-types/media-types.xhtml#font

This fixes a problem when trying to use custom fonts hosted locally with elm reactor, for example using `@font-face`

For example
```css
@font-face {
  font-family: 'Source Code Pro';
  font-style: normal;
  font-weight: 400;
  src: local('Source Code Pro'), local('SourceCodePro-Regular'), url(/source-code-pro.woff2) format('woff2');
}
```
does not work with elm reactor.

When the font file is loaded it will be wrapped in a HTML document by the elm reactor server (at least in 0.19.0).


evancz:
I added them in https://github.com/elm/compiler/commit/536d50c2e0cd5022b8d63e40050289a5107de99f, but with two variations:

1. I kept the list alphabetical by extension. I don't know if that is ideal, but it seems good to stick to it and decide on a different order explicitly.
2. This revealed that there was already an entry for `.ttf` but it looks like it needed to change from `"application/x-font-truetype"` to `"font/ttf"` based on the IANA documents you shared.

Let me know if everything looks alright in the patch, and the changes should be available with the next release of Elm.

PR #1989, opened 2019-10-11
add link to elm slack channel

BKSpurgeon:
**Why the change?**

To make it super easier for beginners to ask for help - by posting the slack channel right there!

Hope this helps.


evancz:
I do not have access to that domain, so I wouldn't want it to go down but be directly encoded in the compiler.

There is also a possibility of moving from Slack to something like Zulip, so at some point I may need to go through the compiler anyway to update about that. The best thing may be to write these messages to point folks to `elm-lang.org/community` instead. Not 100% sure at this time.

PR #1991, opened 2019-10-20
Typo fixing in the hint documents

jfmengels:
I have been through the documents in the `hints/` folder, and I found a few typos/errors and one broken link.

## SSCCE

NA

## Additional Details

I noticed a missing word in the [bad-recursion](https://github.com/elm/compiler/blob/master/hints/bad-recursion.md) but I wasn't sure how to fix, and therefore refrained to do so in the PR. In the following sentence, I think that there is a missing word where I added *FOR*

> A common case is a JSON decoder *FOR* a discussion forums where a comment may have replies, which may have replies, which may have replies, etc.

Let me know if you wish for me to change anything or to take a look at something else.

### DIFF
--- a/hints/bad-recursion.md
+++ b/hints/bad-recursion.md
@@ -10,7 +10,7 @@ There are two problems that will lead you here, both of them pretty tricky:
 
 ## No Mutation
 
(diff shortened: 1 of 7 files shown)


evancz:
Woah, thank you so much for looking through and making these corrections!

PR #1996, opened 2019-10-23
Fix markdown links in installation instructions.

mthadley:
I was passing through and noticed some incomplete markdown links. I assumed that they were still intended to be links and added back the references from a previous commit.

Thanks!


evancz:
Thank you!

PR #2003, opened 2019-10-25
Hint that uglify-js can do multiple passes

HenkPoley:
**Quick Summary:** Talk about the `,passes={n}` addition to uglify-js.

## SSCCE

```elm

```

- **Elm:** ???
- **Browser:** ???
- **Operating System:** ???

## Additional Details

???

### DIFF
--- a/hints/optimize.md
+++ b/hints/optimize.md
@@ -46,7 +46,7 @@ min="elm.min.js"
 
 elm make --optimize --output=$js $@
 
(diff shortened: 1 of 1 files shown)


evancz:
I think it would be good if people knew to choose a higher `passes` if they have the time, but I would not want people to think that `passes=1` is something special and be afraid to change it. Perhaps people would see that there and think that we did some testing to say that this is a good number to use in particular.

So I think the less discoverable approach has the benefit of only emphasizing details that are extremely important to getting good minification. I'm leaning towards that being most important.

PR #2033, opened 2019-12-10
add Travis CI Linux x86_64 build and tags releases

rlefevre:
* Linux binary released only on master tags
* emails sent only on status change (and only to myself for now)
* split-sections added to GHC options
* updated to Alpine 3.11 with ghc 8.6.5

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ cabal-dev
 cabal.sandbox.config
 .DS_Store
 *~
+travis.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+language: minimal
+services: docker
+
+env:
+  global:
+    - LINUX_ARCHIVE=binary-for-linux-64-bit.gz
+
+before_install:
+  - docker build -t elm -f installers/linux/Dockerfile .
+  - docker cp $(docker create elm):/usr/local/bin/elm .
+  - gzip -9 -c elm > $LINUX_ARCHIVE
+
+deploy:
+  provider: releases
+  api_key:
+    secure: Yz2Lo4u9rZQ7Ee7ohAsrZpkqsYDUerCSMdSQIH8ryrf7phHhiloPEkTKsM+NupHqU/LEAVsunxbau4QrCEjA2vPavAPVk8cKomRUWK/YjbXHKa24hPkal2c+A2bnMQ6w3qYk/PjL9rW+Goq++/SNLcYZwHBV0Chl2blivMwWCSA=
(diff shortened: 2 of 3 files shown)


rlefevre:
It's not ready, please wait.

rlefevre:
Now that `Dockerfile` is included in elm sources, I have removed the duplicate `git clone` and `git checkout` and used `docker build` context instead that allows to copy the current directory to the container.

pjhanzlik:
Currently the `Dockerfile` change accidentally deleted ` RUN rm worker/elm.cabal` line which breaks `docker build`. An alternative fix is to replace `COPY . .` with explicit copy instructions:
```Dockerfile
COPY builder/ builder/
COPY compiler/ compiler/
COPY reactor/ reactor/
COPY terminal/ terminal/
COPY LICENSE LICENSE
COPY cabal.config cabal.config
COPY elm.cabal elm.cabal
```

rlefevre:
@pjhanzlik `RUN rm worker/elm.cabal` has been replaced by adding `woker` in `.dockerignore`.
Note that the PR is tested, see [the logs](https://travis-ci.org/elm/compiler/builds/637451417?utm_source=github_status&utm_medium=notification).

I prefer to use an ignore file to an explicit COPY list because this will still work if some commit add some new source code directories without updating the `Dockerfile`.

pjhanzlik:
Given the `Dockerfile` uses `.dockerignore`, maybe there should be a comment telling readers to check it so they know what `COPY . . ` does?  That could be helpful hint if `.dockerignore` and `Dockerfile` get separated sometime in the future.

It might also be worthwhile to explain why `ca-certificate` is a docker build dependency in a comment as well.  Here's an example of [a potential `docker build` issue if `ca-certificate` isn't updated](https://gist.github.com/pjhanzlik/fa8a6964075d319b13a67b4de2559e09).

rlefevre:
@pjhanzlik I have removed ca-certificates because alpine comes by default with them, but Debian doesn't (your example is with Debian).

evancz:
Let's move the `.dockerignore` file into `installers/linux/` to be right next to the `Dockerfile`. Should be good to go from there!

rlefevre:
Actually, it seems that moving the `.dockerignore` requires a recent version of Docker and to enable BUILD_KIT. But thinking more about it, It's likely better to remove it and copy things as @pjhanzlik did.

This requires listing explicitly the needed directories in the `Dockerfile`, but it's more explicit and everything is now in a single place.

evancz:
Great, I like that alternative a lot! Thanks for your patience!

PR #2044, opened 2020-01-04
Add docker configuration files

aminnairi:
**Quick Summary:**

Elm should have a Docker / Docker Compose example configuration for using the binary inside a container. It's not that hard and by providing these files it can become trivial to start getting productive using Elm inside a container. Save some unecessary manual researches.

Plus, it would be good to publish a built image and request an official Elm namespace from the hub.docker.com provider.

## SSCCE

```elm

```

- **Elm:** ???
- **Browser:** ???
- **Operating System:** ???

## Additional Details

???

### DIFF
--- a/docs/docker/Dockerfile
+++ b/docs/docker/Dockerfile
@@ -1,0 +1,23 @@
+# Lightweight GNU/Linux distribution for a lightweight image
+FROM alpine:latest
+
+# Installing the necessary tools to download the Elm binary from GitHub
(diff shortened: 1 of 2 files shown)


evancz:
I don't think the core team should take on maintenance of more install paths. There are lots of ways to mess up a release, and I think it's safer to just maintain the "install on Linux" instructions than an additional Docker thing that needs to be kept in sync and tested. I think it could work as a community thing though!

PR #2058, opened 2020-02-05
remove unneeded slash characters from Dockerfile

rlefevre:
Sorry, missed that in previous PR (this has no functional impact, it's just cleaner because there was a slash for 3 directories, but not the 4th one, which was unintentional but looked like it was).


rlefevre:
Sorry, missed that in previous PR.

Issue #1972, opened 2019-08-20
Elm reactor doesn't serve large files correctly


mgree:
**Quick Summary:** `elm reactor` doesn't properly serve a wasm file (magic numbers corrupted; downloaded binary files differ in size).

Seems to be the same as https://github.com/elm-lang/elm-reactor/issues/214.

## SSCCE

Put z3w.wasm from https://github.com/cpitclaudel/z3.wasm/releases/tag/v0.1 in a directory and serve it from `elm reactor`. Download z3w.wasm---the file sizes differ. Doing the same procedure with `python -m http.server` yields a file of correct size.

```
$ ls -al | grep z3w.wasm
-rw-r--r--@  1 mgree  staff  18267869 Aug 20 14:59 elm.z3w.wasm
-rw-r--r--@  1 mgree  staff  18267216 Aug 20 14:42 py.z3w.wasm
-rw-r--r--@  1 mgree  staff  18267216 Aug 20 14:42 z3w.wasm
```

- **Elm:** 0.19.0
- **Browser:** Firefox 68.0.1 and Chrome 76.0.3809.100
- **Operating System:** OS X 10.13.6

## Probable fix

The reactor seems to be sending the wasm file as text/html rather than application/octet-stream:

```
$ wget -O elm.z3w.wasm http://localhost:8000/app/z3w.wasm
--2019-08-20 14:59:51--  http://localhost:8000/app/z3w.wasm
Resolving localhost (localhost)... 127.0.0.1, ::1, fe80::1
Connecting to localhost (localhost)|127.0.0.1|:8000... connected.
HTTP request sent, awaiting response... 200 OK
Length: unspecified [text/html]
Saving to: ‘elm.z3w.wasm’
```

I suspect that text munging (e.g., null bytes, newlines, carriage returns, etc.) is the source of the problem. Doing better Content-Type inference would help (see also https://github.com/elm-lang/elm-reactor/issues/217).

Anton-4:
I have a similar problem with a csv file. My csv file content is wrapped with:

```
<!DOCTYPE HTML>
<html>

<head>
	<meta charset="UTF-8">
	<title>~/data/category-brands.csv</title>
	<style type="text/css">
		@import url(/_elm/source-code-pro.ttf);

		html,
		head,
		body,
		pre {
			margin: 0;
			height: 100%;
		}

		body {
			font-family: "Source Code Pro", monospace;
		}
	</style>
	<link type="text/css" rel="stylesheet"
[...]

PR #2059, opened 2020-02-07
Added many more MIME types

Anton-4:
should fix some file contents being erroneously wrapped in HTML when using reactor #1972

**Quick Summary:** Reactor served some file contents wrapped in html. This occurred for me when using csv files and made it difficult to use reactor for my application. Others have experienced this problem as well for other file types, see #1972 .
I added many mime types so certain files would no longer be returned as html. I did not yet test my new code because I could not find instructions on how to build elm. I will gladly perform testing given the necessary instructions

## SSCCE

1. In an elm project, put any csv file in assets folder
2. `elm reactor`
3. `wget http://localhost:8000/FILE_NAME.csv` will return the file contents with lines of html added.

- **Elm:** 0.19.1
- **Browser:** firefox 72.0.1 (64-bit)
- **Operating System:** KDE Neon 5.17 > Ubuntu 18.04


gcoakes:
I'm not a contributor to this project or anything, but @Anton-4 , it looks like you can compile elm by running `docker build -t elm -f installers/linux/Dockerfile .` from the project root. That will create a docker image which contains the compiled version of elm at `/usr/local/bin/elm`. It seems like a really weird way to build it when you could also just run `cabal build`, but that's how it's done in the .travis.yml.

Anton-4:
Thanks for the tip, I'll make sure to give it a try!

PR #2076, opened 2020-03-20
Replace uglifyjs with terser

j-maas:
[terser](https://github.com/terser/terser) does not require two passes. I tried out running the `--compress` and `--mangle` passes separately, but running one pass with `--compress --mange` yields the exact same byte count (just as it should be). (uglifyjs, even in the current version 3.8.0, still requires two passes.)

The compression is very close to uglifyjs. uglifyjs takes a 691.340 bytes `elm.js` to 147.094 bytes, while terser yields 149.607 bytes.

The reason I tried out terser was that uglifyjs does not handle modern JavaScript (newer than ES5), which lead to an [issue in elm-pages](https://github.com/dillonkearns/elm-pages/issues/83#issuecomment-599667607). Terser does, is nicely documented, and easily replaces uglifyjs.

Supporting modern JS is not really relevant here right now, since the Elm compiler outputs ES5. But that terser only requires one pass makes explaining it easier.

### DIFF
--- a/hints/optimize.md
+++ b/hints/optimize.md
@@ -15,20 +15,18 @@ Okay, but how do we get those numbers?
 
 Step one is to compile with the `--optimize` flag. This does things like shortening record field names and unboxing values.
 
-Step two is to call `uglifyjs` with a bunch of special flags. The flags unlock optimizations that are unreliable in normal JS code, but because Elm does not have side-effects, they work fine for us!
+Step two is to call `terser` with a bunch of special flags. The flags unlock optimizations that are unreliable in normal JS code, but because Elm does not have side-effects, they work fine for us!
 
 Putting those together, here is how I would optimize `src/Main.elm` with two terminal commands:
 
 ```bash
 elm make src/Main.elm --optimize --output=elm.js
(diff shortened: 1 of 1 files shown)


jhrcek:
The comment you're linking talks about `uglify-es` and is two years old.
[uglify-js](https://github.com/mishoo/UglifyJS2) seems to be very much alive judging by the activity in the repository. I have nothing against your proposed PR, just bringing this up to prevent potential misunderstanding.

Y0hy0h:
Sorry, that is true! I got confused there. I will update the comment. Thanks for the correction!

evancz:
I currently use `swc` for my minification. I know people have done evaluations of N different minifiers to try to find the best balance of speed and quality. I think there could be merit to suggesting a different minifier, but I am not confident that it would be terser in particular. (I know how to set up `swc` since that's what I use, so that's a point in favor of being more likely to be able to keep the docs staying up to date if that's the default. But I don't actually know what is a good modern recommendation.)

lydell:
> I know people have done evaluations of N different minifiers to try to find the best balance of speed and quality.

In case somebody is looking for that, here’s one: https://github.com/lydell/elm-minification-benchmarks

PR #2082, opened 2020-04-10
Add viewport meta tag to HTML page builder

NeilW:
Add a viewport meta tag when generating an HTML page directly, 
so that iPhones don't try and zoom out

**Quick Summary:** 

An old issue https://github.com/elm-lang/elm-make/issues/160

## SSCCE

```elm

```

- **Elm:** 0.19.1
- **Browser:** Safari - iPhone 7
- **Operating System:** 13.4.1


bburdette:
@NeilW Maybe take this change over to Zokka?

PR #2138, opened 2020-09-16
Personalize MISSING EXPRESSION message

landongrindheim:
**Quick Summary:**
I encountered this error message when playing around in elm repl. I had to re-read the sentence a couple times to make sure I understood what the compiler was telling me. I hope this small change makes the message just a little easier to understand 😄

I also updated the link to the ContributorAgreement as I encountered a 404 when I decided to submit this change. Let me know if you'd like that to be in a separate pull request 🙂

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -6,7 +6,7 @@ possible.
 
 ## Licensing
 
-You need to sign the [contributor agreement](ContributorAgreement.pdf)
+You need to sign the [contributor agreement](../ContributorAgreement.pdf)
 and send it to <info@elm-lang.org> before opening your pull request.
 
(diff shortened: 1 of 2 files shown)


landongrindheim:
I'm not sure why this is failing. I tried this earlier (#2137) and had the same result. If you have insight, please let me know.

evancz:
Ah, thank you! Much better with that word filled in!

I have learned that the mere act of using GitHub to do contributions has the contributed code fall under the stated license of the project. I think that wasn't as clear legally in the 90s, so I need to update files here and there to indicate that we are using the [inbound=outbound](https://docs.github.com/en/github/site-policy/github-terms-of-service#6-contributions-under-repository-license) method for this stuff.

PR #2156, opened 2020-12-12
Allow M1 users to continue to install elm via our npm package

loganhenson:
**Quick Summary:**

Allow M1 users to continue to install via npm and use the `x64:64-bit` binary (via rosetta) until we get a native build.

## Additional Details

I would like to continue to allow people with M1 macs to install and use Elm through the npm package, and I think this is a decent stop-gap until we can get a native build. Right now the install fails trying to install a url that does not exist (because the darwin/arm64 combo is not covered), this falls back to the `x64:64-bit` binary download and lets them use Rosetta.

This isn't a problem anywhere else, because the other methods of installation are direct binary downloads, so it asks the user to either install Rosetta or uses Rosetta directly, we just need to make sure we allow them to still get the binary.

Thanks!


evancz:
Thanks to @lydell for pointing me in the right direction on this! And thanks @loganhenson! Seems like a good approach for `npm` folks until GHC can produce native binaries. (Not sure how Mac installers are meant to work at that point though!)

PR #2159, opened 2020-12-22
fix elm build on cabal-install v3.2

fishtreesugar:
Adopt cabal nix-style local builds and add `cabal.project` file for properly build with cabal-install v3.2

Homebrew/homebrew-core#67404

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,0 +1,1 @@
+packages: .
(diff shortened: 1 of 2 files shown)


ahankinson:
NB: This fix is a major hangup for building for homebrew, and it would be nice to be included. 

See: https://github.com/Homebrew/homebrew-core/pull/291551

Perhaps for v1.0?

Issue #2162, opened 2021-01-06
Wrong argument to uglifyjs on .../hints/optimize.md


DrTobe:
**Quick Summary:** The argument `--output=elm.min.js` to uglifyjs should be `--output elm.min.js`

## Additional Details

On https://github.com/elm/compiler/blob/master/hints/optimize.md, this appears twice. First, in "here is how I would optimize", second in the scripts section.

uglifyjs fails with "ERROR: invalid option --output=elm.min.js"

I have installed uglify-js 3.12.4

If you can not reproduce this, please let me know. Thanks.

Eptarch:
I happen to encounter the same.
Inspecting `uglifyjs -h` showed that uglifyjs doesn't know what `--output=elm.min.js` means, but it perfectly understands `--output elm.min.js`
so the .sh should look like this:
```
#!/bin/sh

set -e

js="elm.js"
min="elm.min.js"

elm make --optimize --output=$js $@

uglifyjs $js --compress "pure_funcs=[F2,F3,F4,F5,F6,F7,F8,F9,A2,A3,A4,A5,A6,A7,A8,A9],pure_getters,keep_fargs=false,unsafe_comps,unsafe" | uglifyjs --mangle --output $min

echo "Initial size: $(cat $js | wc -c) bytes  ($js)"
echo "Minified size:$(cat $min | wc -c) bytes  ($min)"
echo "Gzipped size: $(cat $min | gzip -c | wc -c) bytes"
```
I hope it helps ;]

Caleb-Irwin:
I ran into this exact problem some time ago when I was first learning Elm. I eventually figured it out. Is there a PR for this?

I can also confirm this works: 
```
uglifyjs elm.js  --compress "pure_funcs=[F2,F3,F4,F5,F6,F7,F8,F9,A2,A3,A4,A5,A6,A7,A8,A9],pure_getters,keep_fargs=false,unsafe_comps,unsafe" | uglifyjs --mangle --output elm.min.js
```

Caleb-Irwin:
Is this a problem on all systems (or shells) or only some? Would changing it make it not work on some? If not I will open a PR but I am not certain of this. Thoughts?

evancz:
Thanks to @Caleb-Irwin for the fix.

Caleb-Irwin:
No problem! Happy to help Elm.

PR #2167, opened 2021-01-31
Fixed Uglifyjs Command

Caleb-Irwin:
Fixed wrong argument given to Uglifyjs in [hints/optimize](https://github.com/elm/compiler/blob/master/hints/optimize.md). See issue #2162.


evancz:
Looks like they changed whether `=` was allowed on flags in a more recent version. Thank you for the edit!

Issue #2232, opened 2021-11-07
Cross-platform Docker setup fails when using npm install elm


basghar:
**Quick Summary:** ???
We use docker for support multi-platform dev env. On M1 macs, `npm install -g elm` fails because there is no pre-build binary for `linux_arm64`. We want to avoid having different docker for different platform.

```
> node install.js

-- ERROR -----------------------------------------------------------------------

I am detecting that your computer (linux_arm64) may not be compatible with any
of the official pre-built binaries.

I recommend against using the npm installer for your situation. Check out the
alternative installers at https://github.com/elm/compiler/releases/tag/0.19.1
to see if there is something that will work better for you.

From there I recommend asking for guidance on Slack or Discourse to find someone
who can help with your specific situation.

--------------------------------------------------------------------------------
```

- **Elm:** 0.19.1
- **Browser:** N/A
- **Operating System:** Docker on M1 Mac using arm64 linux.

## Additional Details
None

cllns:
Just ran into this same issue.

A workaround I found is to specify `v0.19.4`, which has [a commit that added support for ARM (for macOS Silicon)](https://github.com/elm/compiler/pull/2156). It also added support for `linux_arm64` as a byproduct of this.

The current release, one patch version higher at `v0.19.5` includes [a commit](https://github.com/elm/compiler/commit/759bf04a298a357018d13c85cbb315549893c26d) that continues to support `apple_arm64` but inadvertently **drops** support for `linux_arm64` (which is what Docker uses on hosts that are Apple Silicon, i.e. `apple_arm64`).

akoppela:
As an alternative `nix` can be used to build docker images. Because `nix` has support to build Elm on ARM.

mageekguy:
Hi! Is it possible to merge this PR? Or to say what is the problem with it? Because I need to use a `linux_arm64` version of `elm` with `npm`, and without this fix, it is currently impossible to do that.

gpositive:
Hello,

Any news regarding this PR? It's still a blocker for us.
Is there any way we can help progressing this a bit faster?

Thanks!

lydell:
https://github.com/elm/compiler/pull/2287#issuecomment-1632351816

evancz:
The [0.19.2](https://github.com/elm/compiler/releases/tag/0.19.2) release includes a Linux ARM binary, so this should be possible through `curl` and `npm` now.

PR #2234, opened 2021-12-01
Expand npm installer to include ARM64 Linux

cllns:
A quickfix to #2232 (see comments there), though it may be re-written better or removed entirely (the only main architecture/OS combo this seems to leave out now is ARM Windows).

Or maybe we want to un-do https://github.com/elm/compiler/commit/759bf04a298a357018d13c85cbb315549893c26d in its entirety?

(Note that this lets Apple Silicon users with Elm in docker containers use the npm installer)

### DIFF
--- a/installers/npm/download.js
+++ b/installers/npm/download.js
@@ -21,7 +21,8 @@ module.exports = function(callback)
 		darwin_x64: 'mac-64-bit',
 		darwin_arm64: 'mac-64-bit',
 		win32_x64: 'windows-64-bit',
-		linux_x64: 'linux-64-bit'
+		linux_x64: 'linux-64-bit',
+		linux_arm64: 'linux-64-bit'
 	}[process.platform + '_' + process.arch];
 
 	verifyPlatform(version, platform);


sporto:
Hi @evancz, could we please merge this fix? This is becoming a real issue as more developers are moving to Macs M1 and using Docker.

avh4:
How is a linux-arm64 docker container able to run linux-x64 binaries?  Is that somehow using Rosetta inside the docker container?  Does it only work on MacOS?   Does your docker image have qemu-user-binfmt configured, or can all linux-arm64 docker images somehow run linux-x64 binaries somehow (including on a linux-arm64 host)?

If this working relies on Rosetta and/or qemu-user-binfmt, then I think this is probably incorrect to merge since it will get and try to run binaries for the wrong architecture on machines that don't have emulation services configured.

Also for the record, imo we ought to make official linux-arm64 binaries and then update the npm installer to use those when appropriate.

sporto:
@avh4 We tried downloading the binary for linux manually from releases and run it inside Docker on a M1. It works, but it is very slow. Unsure what is going on, but yes, doesn't seem like the right approach.

lydell:
@sporto Which Docker image do you use? I don’t have an M1, but some colleagues do, and as far as I remember running the Linux x64 binary in `node:16-bullseye` failed for them. If you have time, creating a minimal Dockerfile that demonstrates successfully running Elm on an M1 would be great!

(Btw, [elm-tooling](https://elm-tooling.github.io/elm-tooling-cli) currently does not have any arch checks, so you should be able to use it to install the (only) Linux binary in Docker.)

sporto:
@lydell We used `node:lts-buster`. One of my collegues tried this as I don't have an M1 myself. 

The dockerfile is something like

```
FROM node:lts-buster@sha256:a6c217d7c8f001dc6fc081d55c2dd7fb3fefe871d5aa7be9c0c16bd62bea8e0c

RUN apt-get update

RUN curl -L -o elm.gz https://github.com/elm/compiler/releases/download/0.19.1/binary-for-linux-64-bit.gz \
    && gunzip elm.gz \
    && chmod +x elm \
    && mv elm /usr/local/bin/
```

alshdavid:
Perhaps adding a check to see if an `elm` binary is available before deciding to download would be better. That way people with Rosetta2 and Box64 can manually install the amd64 binary knowing its limitations while the project maintainers can add `darwin-arm64` and `linux-arm64` binaries later

alshdavid:
Hey @evancz, any chance you can chime in here as the change in to [add verification](https://github.com/elm/compiler/commit/759bf04a298a357018d13c85cbb315549893c26d) in the postinstall breaks the `npm install` on my project when compiled in Docker on MacOS / arm Linux.

I'd raise a PR with the changes myself, but from the GitHub pulse it seems this project is no longer maintained?

nicklayb:
@evancz Any way we could get this merged? Of course we can download the binary manually but I came to a point where that doesn't work either. I'm using `create-elm-app` and has `elm` as a npm dependency, so it tries to download the binary anyway

lydell:
@nicklayb This PR has merge conflicts, because the code looks nothing like this anymore since https://github.com/elm/compiler/pull/2287. So it won’t be merged (as-is).

Also, this PR just points to the regular x86 binary for Linux ARM, which is not going to work (in most cases). People with real Linux ARM computers need a Linux ARM binary (as far as I understood from people testing my PR), and people with macOS ARM running in Docker also need a Linux ARM binary. The [@lydell/elm](https://www.npmjs.com/package/@lydell/elm) package exists for this reason. You can force npm to use my package instead, something like this:

```json
{
  "overrides": {
    "elm": "npm:@lydell/elm@0.19.1-14"
  },
  "dependencies": {
    "create-elm-app": "^5.22.0"
  }
}
```

(Btw, create-elm-app isn’t maintained, so you might want to use some other tool like Vite, Parcel 2, Elm Land or elm-watch.)

PR #2247, opened 2022-03-16
Fix in-browser compiler error display

janwirth:
**Quick Summary:** Fix: Error display breaks on long text - the left and right part of the source code is clipped.

## SSCCE

```elm
string< = "..............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................."
```

It looks like this and I can only scroll to the right:
image

- **Elm:** 0.19.1
- **Browser:** Version 99.0.4844.51 (Official Build) (arm64)
- **Operating System:** OSX 12.0.1 (21A559)

### DIFF
--- a/reactor/src/Errors.elm
+++ b/reactor/src/Errors.elm
@@ -57,6 +57,9 @@ viewError error =
         , style "white-space" "pre-wrap"
         , style "background-color" "black"
         , style "padding" "2em"
+        , style "box-sizing" "border-box"
+        , style "width" "100%"
+        , style "overflow" "scroll"
         ]
         (viewErrorHelp error)
     ]


reactor/src/Errors.elm
```diff
@@ -57,6 +57,9 @@ viewError error =
         , style "white-space" "pre-wrap"
         , style "background-color" "black"
         , style "padding" "2em"
+        , style "box-sizing" "border-box"
+        , style "width" "100%"
+        , style "overflow" "scroll"
```
lydell, line 62: Should this use `"auto"` so there’s no unnecessary scrollbar when there is no overflow?

The default on macOS is to draw scrollbars on top of content, and only show it when scrolling. By changing a system setting, you can make macOS behave like Windows and Linux, which is to let the scrollbars take space and always be visible.

The difference between `scroll` and `auto` is that if no scrolling is needed, `scroll` shows ugly disabled scrollbars while `auto` shows no scrollbars at all.

janwirth:
Confirmed working in lamdera-next build offered by @supermario

reactor/src/Errors.elm
janwirth: Good point!

Issue #2283, opened 2022-10-26
Provide a 0.19.1 binary for linux_arm64 to be used in containers on Apple Silicon (M1)


jsynacek:
**Quick Summary:**
Please, provide a binary that can be used in a linux container running on an Apple Silicon (arm64) host.

Currently, the binary asset [binary-for-linux-64-bit.gz](https://github.com/elm/compiler/releases/download/0.19.1/binary-for-linux-64-bit.gz) is usable, but it's really slow and can't be used if elm is specified as a project dependency:
```
<inside the container>

$ which elm
/usr/local/bin/elm

$ npm run build
...
npm ERR! command sh -c node install.js
npm ERR! -- ERROR -----------------------------------------------------------------------
npm ERR! 
npm ERR! I am detecting that your computer (linux_arm64) may not be compatible with any
npm ERR! of the official pre-built binaries.
npm ERR! 
npm ERR! I recommend against using the npm installer for your situation. Check out the
npm ERR! alternative installers at https://github.com/elm/compiler/releases/tag/0.19.1
npm ERR! to see if there is something that will work better for you.
npm ERR! 
npm ERR! From there I recommend asking for guidance on Slack or Discourse to find someone
npm ERR! who can help with your specific situation.
npm ERR! 
npm ERR! --------------------------------------------------------------------------------
```

## SSCCE

- **Elm:** 0.19.1
- **Browser:** N/A
- **Operating System:** MacOS arm64 + Ubuntu 20.04 container

## Additional Details

See also https://github.com/elm/compiler/pull/2234.

alshdavid:
I just tried to compile the project using docker on a Debian arm64 VM (M1 MacBook) with no luck. I am not familiar with Haskell so I wasn't able to do more

bburdette:
Have you tried installing via nix?  That has worked for me on arm64 (pinebook) but not specifically an M1.

supermario:
@jsynacek if you've not seen it already, I built linux_arm64 binaries and Simon has made them available with `@lydell/elm`: https://github.com/elm/compiler/pull/2287

alshdavid:
Any updates to this? :)

supermario:
@alshdavid you can find linux-arm64 binaries here: https://github.com/lydell/compiler/releases/tag/0.19.1

alshdavid:
Is there a way to tell the install script for the `elm` package not to download the binaries?

https://github.com/elm/compiler/blob/master/installers/npm/download.js#L27

[...]

Issue #2007, opened 2019-10-28
What are use-cases for 32-bit binaries?


evancz:
**Quick Summary:** Some people want to use Elm on 32-bit machines.

This meta issue is meant to gather relevant requests to get an idea of what the situation is where folks are using that kind of computer.

- https://github.com/elm/compiler/issues/1878
- https://github.com/elm/compiler/issues/1962

evancz:
The closest thing we have to "build from source" instructions is the Dockerfile for creating our portable linux binary. It only aims to work on a specific version of Alpine Linux, but perhaps it can be useful to people trying to build binaries for other platforms in the meantime.

https://github.com/elm/compiler/blob/master/installers/linux/Dockerfile

PR #2287, opened 2022-11-19
Free the npm package from third party dependencies

### LINKED ISSUE #2229: npm: Change request to other packages
The npm package `request` is under maintenance mode: https://github.com/request/request/issues/3142
Therefore, you need to change `request` to other packages.

Alternative libraries to request: https://github.com/request/request/issues/3143

lydell:
# Status update (2023-09-01)

Me and Evan talked on Slack, and decided this is not happening after all. Not at this time (possibly never). Evan wanted to spend time on paternal leave and compiler exploration, rather than reviewing and tweaking npm package and build stuff.

However, I’ve rebranded the `@lydell/elm` package as an “unofficial elm package” (instead of an “experiment”) and that package will continue to exist, so feel free to use that!

```
npm install @lydell/elm
```

If you use elm-webpack-loader, which depends on the `elm` npm package, you can use the following package.json to override `elm`to `@lydell/elm`:

```json
{
  "overrides": {
    "elm": "npm:@lydell/elm@0.19.1-12"
  },
  "dependencies": {
    "elm-webpack-loader": "^8.0.0"
  }
}
```

# Linked issues

- Closes #2229.
- Related to #2007.
- Related to #2283.
- Related to #2232.

# Background

Me, @supermario and @evancz (with the help of some more folks) have been working on a new version of the `elm` npm package.

- Mario has compiled Elm for Linux ARM: https://github.com/supermario/elm-tooling-compiler/compare/distribution. (Linux ARM binaries are important for Docker on macOS ARM, as well as actual Linux ARM machines.)
- I have:
    - Added support for ARM binaries in the `elm` npm package.
    - Improved how the binaries are distributed for npm.

We have successfully tested all of the above using a temporary npm package (`@lydell/elm`): https://discourse.elm-lang.org/t/help-test-the-new-npm-elm-package/8761. elm-format is also successfully using this npm package technique: https://github.com/avh4/elm-format/pull/781

This PR is the result of my work. If you want to avoid getting into Linux ARM compilation, it’s easy to remove the Linux ARM package.

# The current elm npm package

The current `elm` npm package downloads the Elm binary for the current platform using the [request](https://www.npmjs.com/package/request) package. If there is no binary for the current platform, it exits with a helpful message.

Upsides of this approach:

- Downloading only the needed binary instead of shipping all of them uses less bandwidth and disk space. It saves 25-50 MB per platform.

Downsides of this approach:

- `request` is deprecated.
- Installing `request` results in 47 packages in `node_modules`, weighing in at 5 MB.
- The Elm binary is downloaded on every install, even if the `elm` npm package is cached.

The reason for using `request` rather than Node.js’ `https` module directly, is because `request` supports the `HTTPS_PROXY` environment variable, which is important to people behind corporate proxies or firewalls.

# The new elm npm package

This PR switches to a different technique for distributing the binaries.

- There is one _main npm package_ called `elm`.
- Then there is one _binary npm package_ for each platform, called for example `@evancz/elm_darwin_arm64`.

The binary packages declare which OS and CPU they are compatible with. For example:

```json
  "os": [ "darwin" ],
  "cpu": [ "arm64" ]
```

The main npm package depend on the binary packages via [optional dependencies](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#optionaldependencies):

```json
    "@evancz/elm_darwin_arm64": "0.19.1-0",
    "@evancz/elm_darwin_x64": "0.19.1-0",
    "@evancz/elm_linux_arm64": "0.19.1-0",
    ...
```

When installing, `npm` fetches the metadata for all the optional dependencies and only installs the one with a matching OS and CPU. If none of them match, `npm` still considers the install successful. The main npm package still contains an install script that gives a helpful error.

Upsides of this approach:

- No deprecated dependencies.
- No third party dependencies.
- Same bandwidth and disk space savings as before – and even more actually, since we got rid of the 5 MB of the `request` package.
- Faster, since there are fewer requests to the npm registry.
- The binaries go into the regular npm cache.
- Security: `package-lock.json` contains SHA-512 hashes of every package. So once installed, you can be sure that you get exactly the same Elm binary every time. The current npm package downloads the binaries from GitHub Releases without checking what it got. GitHub Releases is mutable, while the npm registry is immutable.
- One tester of `@lydell/elm` reported that the new package works behind a corporate firewall while the old one does not: https://discourse.elm-lang.org/t/help-test-the-new-npm-elm-package/8761/11?u=lydell
- People behind corporate proxies only need to configure npm, not also whatever the elm npm package uses to download binaries. Check out [this document](https://docs.npmjs.com/cli/using-npm/config/) for more information on how environment variables like [NO_PROXY](https://docs.npmjs.com/cli/using-npm/config#noproxy), [HTTP_PROXY](https://docs.npmjs.com/cli/using-npm/config#proxy), and [HTTPS_PROXY](https://docs.npmjs.com/cli/using-npm/config#https-proxy) are handled by npm.

Downsides of this approach:

Practically, none. No issues were reported during the [test run of `@lydell/elm`](https://discourse.elm-lang.org/t/help-test-the-new-npm-elm-package/8761), and none have been reported during the 9 months of usage of that package since then either. For completeness, here are the factual differences to the current npm package:

- The `--omit=optional` (or `--no-optional`) npm flag can no longer be used. It tells npm to skip all `"optionalDependencies"`. But that is used by the new `elm` npm package to install the correct binary for each platform.

- npm 9.x, 8.x and 7.x all behave correctly, as does Yarn 1.x and pnpm 7.x and >=6.14.0. In short: Installing the new `elm` npm package will still work with old, unsupported package managers, but is non-ideal. The solution is to upgrade your package manager.

  Here are some notes about older (unsupported) npm versions:

    - `npm@6` prints a “warning” for every binary package that isn’t used on the current platform:

        ```
        npm WARN optional SKIPPING OPTIONAL DEPENDENCY: @lydell/elm_darwin_x64@0.19.1-1 (node_modules/@lydell/elm/node_modules/@lydell/elm_darwin_x64):
        npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for @lydell/elm_darwin_x64@0.19.1-1: wanted {"os":"darwin","arch":"x64"} (current: {"os":"darwin","arch":"arm64"})
        npm WARN optional SKIPPING OPTIONAL DEPENDENCY: @lydell/elm_linux_x64@0.19.1-0 (node_modules/@lydell/elm/node_modules/@lydell/elm_linux_x64):
        npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for @lydell/elm_linux_x64@0.19.1-0: wanted {"os":"linux","arch":"x64"} (current: {"os":"darwin","arch":"arm64"})
        npm WARN optional SKIPPING OPTIONAL DEPENDENCY: @lydell/elm_win32_x64@0.19.1-0 (node_modules/@lydell/elm/node_modules/@lydell/elm_win32_x64):
        npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for @lydell/elm_win32_x64@0.19.1-0: wanted {"os":"win32","arch":"x64"} (current: {"os":"darwin","arch":"arm64"})
        npm WARN optional SKIPPING OPTIONAL DEPENDENCY: @lydell/elm_linux_arm64@0.19.1-2 (node_modules/@lydell/elm/node_modules/@lydell/elm_linux_arm64):
        npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for @lydell/elm_linux_arm64@0.19.1-2: wanted {"os":"linux","arch":"arm64"} (current: {"os":"darwin","arch":"arm64"})
        ```

      That’s noisy but harmless.

    - `npm@7.0.0` prints no warnings but installs _all_ optional dependencies. That’s wasteful but harmless. The latest 7.x works correctly though.

    - Yarn 2+ (“Berry”) is now supported (the current `elm` npm package does not support that). Yarn 2+ _always_ executes the `"bin"` of a package using `node` (regardless of [installation mode](https://yarnpkg.com/configuration/yarnrc#nodeLinker): pnp, pnpm, node-modules). Luckily, it’s easy to detect Yarn 2+ and do the same bailout as we do on Windows anyway. Note: If a user is on macOS or Linux, installs Elm using Yarn 2+, and executes `elm` often via their editor, then it’s an anti-pattern to use Yarn 2+ since every `elm` invocation has to pay for the `node` startup cost as well. But that’s up to Yarn to fix: https://github.com/yarnpkg/berry/issues/882

# What stayed the same

In short: Most things. The main difference is that the new `elm` npm package gets the Elm binary from other npm packages instead of from GitHub Releases.

- We still need a `bin/elm` written in Node.js. The binary packages _can_ technically have `"bin"` fields in their package.json files, but it’s not reliable – and does not work in pnpm by design. So there needs to be a `"bin"` field in the main npm package. And that needs to point to something in the main npm package itself. So, a Node.js file, which finds the correct binary for the current platform and executes it using `child_process.spawn`.
- We still have an install script. The current `elm` npm package downloads the Elm binary on install. And on top of that, on Linux and Windows it downloads it to the path of the `bin/elm` Node.js file, so that you would be invoking the Elm binary directly with no overhead. The new `elm` npm package also does that optimization on install (with hard links). (On Windows, npm forces executing `"bin"` with Node.js so that optimization is not possible.)
- We still optimize on first run. The current `elm` npm package downloads the Elm binary on the first run, if `--ignore-scripts` was used during install. The new `elm` npm package does the above optimization on the first run if `--ignore-scripts` was used.

# Diff tip

I renamed `download.js` to `binary.js` and then edited it. By default, `git` considers a file to be renamed if there’s at least 50 % similarity. However, `binary.js` is less than 50 % similar to `download.js` so the GitHub diff shows `download.js` as deleted and `binary.js` as a new file.

Locally you can run:

```
git diff --merge-base master --find-renames=30%
```

And then type `/download.js` and press <kbd>Enter</kbd> to jump to the relevant part. I think that diff helps to see that the new approach isn’t all that different from before.

# Things left out from this PR

The approach of this PR was inspired by esbuild:

- https://github.com/evanw/esbuild/pull/1621
- https://github.com/evanw/esbuild/blob/ecc9eeb5ec13cf32e0690e92aa2e933d1267beb4/lib/npm/node-platform.ts
- https://github.com/evanw/esbuild/blob/ecc9eeb5ec13cf32e0690e92aa2e933d1267beb4/lib/npm/node-install.ts

If you read through those files, you’ll see that they are much more elaborate than `binary.js` in this PR.

- esbuild tries to detect if `--omit=optional` was used, and if so try to manually download the binary first from the npm registry, then from GitHub. It’s unclear why that complexity would be necessary. This PR instead shows a helpful error message in this situation instead. Nobody complained about this when testing the `@lydell/elm` package.
- esbuild supports the `ESBUILD_BINARY_PATH` which basically bypasses its installer and turns its `"bin"` into a script that executes the binary at `ESBUILD_BINARY_PATH`. Something like this was mentioned in a discussion in `#core-coordination` on Slack. Me and Evan discussed it, and decided to leave it out since it wasn’t as clear-cut as one might first think. (More details in the aside below.)

<details>

<summary>Aside about unsupported platforms</summary>

@alshdavid posted in `#core-coordination` on Slack, which got the ball rolling on this work:

> Hi Elm team, David here. I am a contributor of Parcel and we are having the issue where the recently introduced `verifyPlatform()` function in elm/compiler is preventing the Parcel monorepo (contributors) and Parcel+Elm (consumers) from being `npm install`ed when running on `linux-arm64` devices.
>
> This is an increasingly common use-case with Parcel contributors and consumers due to the rise of M1 MacBooks (Docker & VMs) and Linux ARM hosts being used more frequently as build agents.
>
> **For contributors of Parcel:** The Parcel monorepo fails to `npm install` at all on `linux-arm64` hosts, which is pretty critical given the use of Docker in development.
>
> **For consumers of Parcel and Elm:** Projects using Parcel and Elm and are building on `linux-arm64` machines are unable to `npm install`/build their projects. This happens locally when developers are using Parcel+Elm in Docker on an M1 MacBook and happens on build agents, but there is increasing interest in arm based devices so we anticipate this will only become more common.
>
> I'm happy to raise some PRs to address these issues. Given this is a pretty serious blocker, would there be someone available to review my work as I raise PRs?
>
> **Just spitballing some ideas:**
>
> I was thinking of initially introducing an env variable (e.g. `ELM_BINARY_LOCATION=/usr/local/bin/elm`) which, if set, will skip the download and platform verification.
>
> This would allow consumers who are on `linux-arm64` devices and using a translation layer like Box64 or Rosetta2 to use the `amd64` version while being aware of the limitations. It's also an extensible solution given it might work for `win-arm64` and other platforms with translation layers and without prescribed binaries.
>
> Secondly, I was thinking of adding some GitHub actions to automate the compilation of binaries for the various platforms and produce a GitHub release with those binaries.

Supporting Linux ARM solves their immediate problem, so the discussion becomes more about a hypothetical future scenario which could happen as a new currently unsupported platform becomes popular.

What happens then is that `npm install` fails due to the `elm` npm package erroring by design due to the unsupported platform. This fails the entire `npm install`, not just the install of the `elm` package.

In Parcel’s case, it means contributors can’t install the project even if they didn’t plan to work on the Elm parts of Parcel.

Here are some workarounds:

- `npm install --ignore-scripts`. That will skip our install script, so no failure there. (If you then try to run `npx elm`, it’ll fail with the same message as the install script.) However, if you’re unlucky there’s some other dependency that does require install scripts.
- `npm install --legacy-peer-deps` can help if `elm` is a peer dependency. For example, [elm-webpack-loader has `elm` as a peer dependency](https://github.com/elm-community/elm-webpack-loader/blob/a597c3f06711324318a2247a394001fdad4f1270/package.json#L36). That will skip automatic installation of “peer dependencies”, which will then skip installing the `elm` npm package. However, there might be other peer dependencies of the project that you now need to find and install yourself. Also, this only works if elm is actually specified as a peer dependency. If the project had been using [elm-land](https://elm.land/) instead of webpack, you’re out of luck: It has a [hard dependency on the elm npm package](https://github.com/elm-land/elm-land/blob/bad0e43140cd1da61b17344725d2c6429ebd4aad/projects/cli/package.json#L39).

Here are some ideas we discussed for handling this situation:

- Allow disabling the install script via an env var.
- Allow also pointing to your elm binary via that env var.
- Have `npm install` succeed on unsupported platforms, but `npx elm` (executing the installed elm “binary”) fail with the current error message.
- Remove the install script completely. The first run will do the platform check and optimization for Linux and macOS anyway.
- Parcel could release their own npm package, like `@parcel/elm`. Note: The only need to replace the main `elm` npm package, and can still re-use the binary packages (like `@evancz/elm_darwin_arm64`). **Note:** This one is doable without any changes to the `elm` npm packages (unlike the above points).

For people running `npm install elm` themselves it’s ideal if it fails during installation on unsupported platforms. They should know about the problem, have a nice experience, and the npm package should be as simple as possible. From the perspective of Parcel, they want it to fail silently, maybe only when someone tries to run elm at some point or whatever. It’s not clear those two use cases are incompatible within a single package.

We decided we just want to keep it simple and not make any design changes like the env var business.

</details>


installers/npm/bin/elm
```diff
@@ -1,54 +1,28 @@
...
-{
-	// Need double quotes and { shell: true } when there are spaces in the path on windows:
-	// https://github.com/nodejs/node/issues/7367#issuecomment-229721296
-	child_process
-		.spawn('"' + binaryPath + '"', process.argv.slice(2), { stdio: 'inherit', shell: true })
-		.on('exit', process.exit);
-}
+var binaryPath = require('../binary.js')();
+child_process
+	.spawn(binaryPath, process.argv.slice(2), { stdio: 'inherit' })
```
lydell, line 27: The old code:

```js
// Need double quotes and { shell: true } when there are spaces in the path on windows:
// https://github.com/nodejs/node/issues/7367#issuecomment-229721296
child_process
  .spawn('"' + binaryPath + '"', process.argv.slice(2), { stdio: 'inherit', shell: true })
```

There is no need to use `shell: true` when executing a .exe file. And when not using `shell: true`, we don’t have to worry about quoting.

I’ve tested this in a folder with spaces on Windows myself. I learned about how this works when working on the [elm-tooling](https://elm-tooling.github.io/elm-tooling-cli/) CLI.

installers/npm/package.json
```diff
@@ -2,7 +2,6 @@
   "name": "elm",
   "version": "0.19.1-5",
   "description": "Installer for Elm: just downloads the binary into node_modules",
-  "preferGlobal": true,
```
lydell, line 5: `preferGlobal` has been deprecated for a long time, and is even removed in newer npm versions.

lydell:
For others following along, me and @ryannhg talked about his installation problem on Slack instead of here.

I verified that installing `@lydell/elm` works on all platforms on GitHub Actions here: https://github.com/lydell/elm-github-actions-test/actions/runs/3722282755

cyberglot:
Any updates on this?

lydell:
@cyberglot No, the status update at the top is still the latest news!

alshdavid:
Does that mean the `elm` package is now officially unmaintained and development has moved to `@lydell/elm`?

supermario:
@alshdavid no it does not mean that, it just means that the community has worked together to provide an unofficial binary for the missing platform until Evan has the time to prioritise this.

For example Evan committed support for the M1 arm binaries last week: https://github.com/elm/compiler/commit/ad9d9d42c4f623c2903ecf35daefaff011075977

lydell:
@Zeneixe Me and supermario collaborated on this: He did builds for more platforms, while I did the npm package. The idea was first to ship an Linux ARM64 binary too. The Linux ARM64 uses pretty complicated and slow Docker build setup, and as far as I remember Evan didn’t want to dive deep into understanding how that works, so it was removed (though not mentioned in the commit message). You can still use my `@lydell/elm` package, though – it contains the a Linux ARM64 binary. (That’s the reason my package is still around.)

PR #2306, opened 2023-07-14
Fix `npm audit` complaining that elm is still using request

cyberglot:
**Quick Summary:**
request has been deprecated, and `npm audit` complains about it. This PR simply removes request and replaces it with axios, semantics are equivalent.

**Additional Info:**
I understand that we have the #2287 PR to overhaul the current npm installation, but it seems that progress there has been extremely slow.

### DIFF
--- a/installers/npm/download.js
+++ b/installers/npm/download.js
@@ -1,7 +1,7 @@
 var fs = require('fs');
 var package = require('./package.json');
 var path = require('path');
-var request = require('request');
+var request = require('axios');
 var zlib = require('zlib');
 
 
@@ -53,7 +53,7 @@ module.exports = function(callback)
 	});
 
 	// put it all together
-	request(url).on('error', reportDownloadFailure).pipe(gunzip).pipe(write);
+	request.get(url).then(gunzip).then(write).catch(reportDownloadFailure);
 }
 
 
--- a/installers/npm/package-lock.json
+++ b/installers/npm/package-lock.json
@@ -1,0 +1,113 @@
+{
+  "name": "elm",
+  "version": "0.19.1-5",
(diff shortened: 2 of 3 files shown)


installers/npm/download.js
```diff
@@ -53,7 +53,7 @@ module.exports = function(callback)
 	});
 
 	// put it all together
-	request(url).on('error', reportDownloadFailure).pipe(gunzip).pipe(write);
+	request.get(url).then(gunzip).then(write).catch(reportDownloadFailure);
```
lydell, line 56: When I run `node install.js` on master, it replaces `bin/elm` (a Node.js script) with an elm binary, which is the expected behavior.

On your branch, `node install.js` results in `bin/elm` becoming an empty file.

`.pipe` is for chaining streams, while `.then` is for chaining promises?

lydell (commented):
It’s good that https://github.com/axios/axios mentions a `http_proxy` environment variable! Because that was one of the reasons for using `request` instead of plain Node.js in the first place.

However, for this to be merged I guess testing / “proof” that there are no proxy stuff regressions compared to the `request` version would be needed.

mather:
I tried implementing it using [https](https://nodejs.org/api/https.html) from the Node.js standard library, but the elm download URL needs to follow a redirect, and it seems to work better if implemented with the [follow-redirect](https://www.npmjs.com/package/follow-redirects) library instead.

My experimental implementation is here: https://github.com/mather/elm-compiler/blob/master/installers/npm/download.js

lydell:
@mather See https://github.com/elm/compiler/commit/41ec49ed921a2409afda483eb9e29197e262fe27

PR #2307, opened 2023-08-18
npm installer: Download elm package without deprecated libraries.

mather:
**Quick Summary:**
[request](https://www.npmjs.com/package/request) is deprecated, and vulnerabilities have been reported in its dependency libraries.
I replaced download feature using following libraries:
- [follow-redirects](https://www.npmjs.com/package/follow-redirects)
- [proxy-from-env](https://www.npmjs.com/package/proxy-from-env)

Ref: https://github.com/elm/compiler/commit/41ec49ed921a2409afda483eb9e29197e262fe27

**Thought:** 
[axios](https://www.npmjs.com/package/axios) also depends on the same libraries (follow-redirects, proxy-from-env), but there is no need to use Promise-based features.
To improve maintainability, I believe that we should adopt libraries with simple enough functions to meet our needs,rather than relying on a library with many additional functions.

### DIFF
--- a/installers/npm/download.js
+++ b/installers/npm/download.js
@@ -2,7 +2,7 @@ var fs = require('fs');
 var path = require('path');
 var zlib = require('zlib');
 var { https } = require('follow-redirects');
-var { ProxyAgent } = require('proxy-agent');
+var { getProxyForUrl } = require('proxy-from-env');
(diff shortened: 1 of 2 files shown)


mather:
I want to review the behavior of proxy and organize the commits, so I close PR.

PR #2330, opened 2024-07-17
add occurs checks for all unifications

### LINKED ISSUE #2241: `elm make` hang then runaway memory usage from orphan process
**Quick Summary:** Compilation hangs and an elm process remains active after `<ctrl-c>`. Memory usage of this process reaches >16GB (memory available to system) which bogged down my system.

## SSCCE

```elm
module Main exposing (..)

foldMap : (a -> b) -> (a -> c) -> (b -> c -> c) -> a -> c
foldMap fab fac fbc a = fac a

break : Float -> (Float, Float)
break input = foldMap identity (\x -> (x, x)) (\( low, high ) x -> ( min low x, max high x )) input
```

- **Elm:** Elm 0.19.1
- **Browser:** NA
- **Operating System:** macOS 11.4

## Additional Details

`elm make` hangs after printing `Compiling ...`. The orphan elm process only continues if waiting for longer than 5-6 seconds before using <ctl-c> to cancel the elm make command.

When changing `(\( low, high ) x ->` to `(\x ( low, high ) ->` the snippet compiles as expected, and changing `( min low x, max high x )` to `( x, x )` results in the compiler error `` `foldMap` needs the 4th argument to be:    ( ∞, ∞ )``.

Came across this writing a foldMap to calculate the min and max of a List Float but have simplified to get a minimal reproduction.

Love your work!

changlinli:
This fixes https://github.com/elm/compiler/issues/2241

Preliminary testing seems to indicate this isn't as bad of a slowdown as it may seem. For more details see https://github.com/Zokka-Dev/zokka-compiler/pull/20

This is meant as a reference PR for other people working on the Elm compiler.

### DIFF
--- a/compiler/src/Type/Unify.hs
+++ b/compiler/src/Type/Unify.hs
@@ -166,15 +166,20 @@ fresh (Context _ (Descriptor _ rank1 _ _) _ (Descriptor _ rank2 _ _)) content =
 guardedUnify :: Variable -> Variable -> Unify ()
 guardedUnify left right =
   Unify $ \vars ok err ->
-    do  equivalent <- UF.equivalent left right
-        if equivalent
-          then ok vars ()
+    do  occursLeft <- Occurs.occurs left
+        occursRight <- Occurs.occurs right
+        if occursLeft || occursRight
+          then err vars ()
           else
-            do  leftDesc <- UF.get left
-                rightDesc <- UF.get right
-                case actuallyUnify (Context left leftDesc right rightDesc) of
-                  Unify k ->
-                    k vars ok err
+            equivalent <- UF.equivalent left right
+            if equivalent
+              then ok vars ()
+              else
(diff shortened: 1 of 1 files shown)


evancz:
Based on my understand of [The Essence of ML Type Inference](https://gallium.inria.fr/~fpottier/publis/emlti-final.pdf), this approach is too heavy handed. This particular section of the chapter seems to confirm as much:

> Unfortunately, many implementations of type inference for ML-the-programming-language do not behave as efficiently as the algorithm presented here. Some spend an excessive amount of time in computing the set of nongeneralizable type variables; some do not treat types as dags, thus losing precious sharing information; **others perform the expensive occurs check after every unification step**, rather than only once at every let construct, as suggested here (S-Pop-Let).

I believe the proper fix is something else, though I am not 100% sure exactly where we are diverging from the algorithm they describe.

PR #2356, opened 2026-06-30
Update README.md

mehryar:
Fixing typo in install instructions.
I'm sure there is a 0.19.2 version in the works but this PR is to prevent people from wasting time troubleshooting why these install instructions aren't working - since there is no 0.19.2 release on GitHub yet.


evancz:
This PR came in right as 0.19.2 was in pre-release, so the information should be correct now.

I think the root is that the links on various websites point to the "latest" version of the file, so this is a consequence of bringing code into the `main` branch before the final public release.

PR #2359, opened 2026-07-08
Fix off-by-one in source regions emitted to JSON and Debug.todo

### LINKED ISSUE #2358: 0.19.2: source regions in --report=json and Debug.todo are 0-based (off by one)
**Quick Summary:** In 0.19.2 the source regions emitted by `elm make --report=json` and by `Debug.todo` / `Debug.todoCase` are **0-based**, so every line and column is one too low. The human-readable terminal report is unaffected.

`Reporting.Annotation.Position` became a packed `Word64#` in 0.19.2, and the module comment states the new contract:

```haskell
-- It is stored as (y,x) so (pos + 1) increments x, and (pos + 0x100000000)
-- increments y. The initial value is (0,0) so conversion is needed to get
-- "editor coordinates" which start at (1,1)
data Position = Position Word64#
```

`Reporting.Render.Code` performs that conversion (`show (row + 1)`), which is why terminal output still looks right. The two places that emit machine-readable coordinates call `A.toRowCol` raw and never convert:

- `compiler/src/Reporting/Error.hs` — `encodeRegion`, used by `--report=json`
- `compiler/src/Generate/JavaScript/Expression.hs` — `regionToJsExpr`, used for `Debug.todo` / `Debug.todoCase`

This matters most for editor tooling (`elm-language-server` and friends consume `--report=json`), which now highlights one line up and one column left.

## SSCCE

**(1) `Debug.todo` reports the wrong line.**

```elm
module Main exposing (main)

import Platform


main : Program () () ()
main =
    Platform.worker
        { init = \_ -> ( thing, Cmd.none )
        , update = \_ m -> ( m, Cmd.none )
        , subscriptions = \_ -> Sub.none
        }


thing : ()
thing =
    Debug.todo "boom"
```

`Debug.todo` is on **line 17, column 5**. Compile with `elm make src/Main.elm --output=out.js` and run it in node:

| | emitted region | runtime message |
|---|---|---|
| 0.19.1 | `{start: {line: 17, column: 5}, end: {line: 17, column: 15}}` | ``TODO in module `Main` on line 17`` |
| 0.19.2 | `{start: {line: 16, column: 4}, end: {line: 16, column: 14}}` | ``TODO in module `Main` on line 16`` |

**(2) `--report=json` reports the wrong region.**

```elm
module M exposing (bad)


bad : Int
bad =
    "not an int"
```

The offending string literal spans **line 6, columns 5–17**. Run `elm make src/M.elm --report=json`:

| | `region` in the JSON | terminal report |
|---|---|---|
| 0.19.1 | `start {line: 6, column: 5}`, `end {line: 6, column: 17}` | `6\|     "not an int"` ✅ |
| 0.19.2 | `start {line: 5, column: 4}`, `end {line: 5, column: 16}` | `6\|     "not an int"` ✅ |

So the terminal renderer is correct in both, and only the JSON regions regressed.

- **Elm:** 0.19.2 (regression; 0.19.1 is correct). Reproduced with the official `elm-0.19.2-linux-x64` binary.
- **Browser:** n/a (also reproduces under node for the `Debug.todo` case)
- **Operating System:** Linux x86-64

## Additional Details

Every region I checked is off by exactly `(-1, -1)`. In a real 548-module application, all 12 `Debug.todo` regions in the emitted JS differ from the 0.19.1 output by exactly one line and one column; a uniform `+1` reproduces the 0.19.1 bundle byte-for-byte.

`main` (`e20e8c8e0`) is still affected — both files are identical to the `0.19.2` tag.

A possible fix, keeping the conversion explicit and in one place:

```haskell
-- compiler/src/Reporting/Annotation.hs   (and add toEditorRowCol to the export list)

-- Editor coordinates start at (1,1); the packed representation starts at (0,0).
{-# INLINE toEditorRowCol #-}
toEditorRowCol :: Word64# -> (Int, Int)
toEditorRowCol pos =
  ( fromIntegral (toRow pos) + 1
  , fromIntegral (toCol pos) + 1
  )
```

then use `A.toEditorRowCol` instead of `A.toRowCol` in `Reporting.Error.encodeRegion` and `Generate.JavaScript.Expression.regionToJsExpr`.

Worth noting: after that change `toRowCol` has **no remaining callers** — its only two consumers both wanted editor coordinates — so simply making `toRowCol` itself 1-based would also work, and is a smaller diff. `Reporting.Render.Code` uses `toRow` / `toCol` directly and already handles its own conversion, so it is unaffected either way.

anvmn:
Fixes #2358.

In 0.19.2 `Reporting.Annotation.Position` became a packed `Word64#`, and its own comment states the new contract:

```haskell
-- The initial value is (0,0) so conversion is needed to get
-- "editor coordinates" which start at (1,1)
```

`Reporting.Render.Code` performs that conversion (`show (row + 1)`), which is why the terminal error report is still correct. The two places that emit *machine-readable* coordinates call `A.toRowCol` raw and never convert:

- `Reporting.Error.encodeRegion` — used by `elm make --report=json`, which editors and `elm-language-server` consume
- `Generate.JavaScript.Expression.regionToJsExpr` — used for `Debug.todo` / `Debug.todoCase`

So every JSON diagnostic is one line up and one column left, and a `Debug.todo` on line 17 crashes with `` TODO in module `Main` on line 16 ``.

## The change

Add a `toEditorRowCol` helper next to `toRowCol` and use it at the two emit sites. `Reporting.Render.Code` uses `toRow` / `toCol` directly and already does its own conversion, so it is untouched.

Note: after this change `toRowCol` has **no remaining callers** — both of its consumers wanted editor coordinates. Making `toRowCol` itself 1-based would be a smaller diff; I kept a separately named function so the (0,0)-based accessor stays available and the conversion is explicit at the point of use. Happy to switch to whichever you prefer.

## Verification

Built with GHC 9.8.4 (`cabal build exe:elm`, so under the package's `-Wall -Werror`): 137/137 modules, links cleanly.

SSCCE from #2358 — `Debug.todo` on line 17, column 5; type error on line 6, columns 5–17:

| compiler | `Debug.todo` region | runtime message | `--report=json` region |
|---|---|---|---|
| 0.19.1 | `(17,5)–(17,15)` | `` on line 17 `` | `(6,5)–(6,17)` |
| 0.19.2 | `(16,4)–(16,14)` | `` on line 16 `` | `(5,4)–(5,16)` |
| **0.19.2 + this patch** | **`(17,5)–(17,15)`** | **`` on line 17 ``** | **`(6,5)–(6,17)`** |

The patched compiler matches 0.19.1 exactly.

Also checked:

- The human-readable report is unchanged — still `6|     "not an int"` with the caret correctly aligned, i.e. not double-incremented.
- On a real 548-module application, all 12 `Debug.todo` regions emitted into the bundle become identical to the 0.19.1 output (Δline = Δcolumn = 0 for every one), where before the patch every one was off by exactly `(-1, -1)`.

### DIFF
--- a/compiler/src/Generate/JavaScript/Expression.hs
+++ b/compiler/src/Generate/JavaScript/Expression.hs
@@ -309,8 +309,8 @@ regionToJsExpr (A.Region start end) =
     ]
   where
     (===>) n v = (JsName.fromLocal n, v)
-    (sr, sc) = A.toRowCol start
-    (er, ec) = A.toRowCol end
+    (sr, sc) = A.toEditorRowCol start
+    (er, ec) = A.toEditorRowCol end
 
 
 
--- a/compiler/src/Reporting/Annotation.hs
+++ b/compiler/src/Reporting/Annotation.hs
@@ -9,6 +9,7 @@ module Reporting.Annotation
   , merge
   , at
   , toRowCol
+  , toEditorRowCol
   , toRow
   , toCol
   , toRegion
@@ -78,6 +79,17 @@ toRowCol :: Word64# -> (Int, Int)
 toRowCol pos =
   ( fromIntegral (toRow pos)
   , fromIntegral (toCol pos)
+  )
+
+
+-- Editor coordinates start at (1,1); the packed representation starts at (0,0).
+-- Anything that leaves the compiler for a human or a tool must go through this,
+-- not through toRowCol.
+{-# INLINE toEditorRowCol #-}
+toEditorRowCol :: Word64# -> (Int, Int)
+toEditorRowCol pos =
+  ( fromIntegral (toRow pos) + 1
+  , fromIntegral (toCol pos) + 1
   )
 
 
--- a/compiler/src/Reporting/Error.hs
+++ b/compiler/src/Reporting/Error.hs
@@ -200,6 +200,6 @@ encodeRegion (A.Region start end) =
     , "end"   ==> E.object [ "line" ==> E.int er, "column" ==> E.int ec ]
     ]
   where
-    (sr,sc) = A.toRowCol start
-    (er,ec) = A.toRowCol end
+    (sr,sc) = A.toEditorRowCol start
+    (er,ec) = A.toEditorRowCol end
 


evancz:
Thank you! I like the approach of having a separate naming convention for "editor rows" and "editor columns".

Sad I did not catch this in the pre-release, but there should be a 0.19.3 relatively soon.

